PR #12, opened 2016-05-03
Adding descriptions and details for each segment

stuartlynn:
Adding descriptions and details for demographic segments. @talos


talos:
Looks good!  Only change would be to lowercase the `Description` and `Details` keys, but I can take care of that before merge.

PR #19, opened 2016-05-10
adding code for assigning names for x55 demographics

stuartlynn:
Hey @talos double check that this looks ok when you have a sec. We will probably update the names tomorrow morning so just double check the code it's self. 

Cheers

Stuart


talos:
Looking...

talos:
(If I see issues I'll just hotfix this branch)

PR #23, opened 2016-05-13
Fix nonzero nondesc acs

andrewxhill:
only 2 found with the error. now on to the enhancements


tasks/us/census/acs.py
```diff
@@ -645,15 +645,15 @@ def columns(self):
...
+            description='The Gini index, or index of income concentration, is a statistical measure of income inequality ranging from 0 to 1. A measure of 1 indicates perfect inequality, i.e., one household having all the income and rest having none. A measure of 0 indicates perfect equality, i.e., all households having an equal share of income.',
             weight=5,
             aggregate='',
             tags=[censustags['demographics'], tags['income_education_employment']])
         income_per_capita = OBSColumn(
             id='B19301001',
             type='Numeric',
             name='Per Capita Income in the past 12 Months',
-            description='',
+            description='Per capita income is the mean income computed for every man, woman, and child in a particular group. It is derived by dividing the total income of a particular group by the total population.',
```
talos: Wow, the Census's own description of this variable is terrible.  Let's go with this for now but seek clarity on what a "particular group" is and why they don't just say "household or individuals not in households".

PR #82, opened 2016-08-16
changed to carto.com

stvno:
The cartodb.com > carto.com redirect doesn't seem to work in the tasks


talos:
👍

talos:
@stvno I checked in with CARTO systems, and apparently new accounts do not have the cartodb.com alias, so this was expected behavior.  Either way, we should be using the new domain!

PR #84, opened 2016-08-17
thailand district populations

talos:


tasks/th/thaipop.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def populate(self):
+        session = current_session()
+        session.execute('INSERT INTO {output} '
+                        'SELECT the_geom, pop2010, name_2 '
+                        'FROM {input} '.format(
+                            output=self.output().table,
+                            input=self.input()['data'].table
+                        ))
+
```
talos: The empty lines below here should be removed.

tasks/th/thaipop.py
```diff
@@ -0,0 +1,122 @@
...
+            name='Name of District',
+            type='Text',
+            weight=5,
+            targets={the_geom: GEOM_NAME},
+        )
+
+        return OrderedDict([
+            ('the_geom', the_geom),
+            ('pop2010', pop2010),
+            ('name_2', name_2),
```
talos: We can provide a nicer colname for end-users -- just "name" would be easier.

tasks/th/thaipop.py
```diff
@@ -0,0 +1,122 @@
...
+        name_2 = OBSColumn(
+            name='Name of District',
+            type='Text',
+            weight=5,
+            targets={the_geom: GEOM_NAME},
+        )
+
+        return OrderedDict([
+            ('the_geom', the_geom),
+            ('pop2010', pop2010),
```
talos: Since the year is a property of the table rather than the column, this should be "pop" or "total_pop" instead of "pop2010".

tasks/th/thaipop.py
```diff
@@ -0,0 +1,122 @@
...
+    def requires(self):
+        return {
+            'sections': SectionTags(),
+            'subsections': SubsectionTags(),
+        }
+
+    def version(self):
+        return 1
+
+    def columns(self):
```
talos: Is there a column in the original table that looks like a consistent geom identifier (akin to FIPS code)?

tasks/th/thaipop.py
```diff
@@ -0,0 +1,122 @@
+
+# coding: utf-8
+
+# In[ ]:
```
talos: This artifact of the IPython notebook could be removed.

talos:
This looks great! Tested on the ETL box and it works. Just a few minor comments above to resolve, then we should be good to go.

tasks/th/thaipop.py
michellemho: It looks like CCA is a unique code identifier, but there are 924 unique codes whereas there are 928 names. Three of the four missing codes can be explained by three locations (two Songkhla Lakes and Wiang Kao) which have null for population and CCA. There might be a duplicate that would explain the last missing code.

talos:
@michellemho can you check your ACS work out into a new branch (`git checkout -B acs-rentburden`) then push that to GitHub?

Once you've done that, you can checkout this branch again (`git checkout newbranchtest`), rewind it to the last relevant commit (`git reset --hard 571601b`) and move github back to the correct commit (`git push origin -f newbranchtest`)

PR #92, opened 2016-08-29
Stats Canada (NHS and Census Surveys)

aspcanada:


docker-compose.yml
```diff
@@ -9,6 +9,9 @@
                                               # will cause persistence issues!!
       "./observatory-extension:/observatory-extension"
     ],
+    "ports": [
```
talos: Was this needed locally?  We don't want to hardcode this port as it messes with CI.

talos:
This looks good -- Census tasks run, excepting three where the column summaries query ends up being too long, which needs to be fixed elsewhere.  NHS is still running.  Once that's done, I'll test putting the data up on our test server.

PR #105, opened 2016-10-03
Frenchdata

michellemho:
adding ETL for French census data


tasks/fr/insee.py
```diff
@@ -0,0 +1,333 @@
...
+            for table_theme in topics:
+                yield FranceCensus(table_theme=table_theme, resolution=resolution)
+
+
+class AllGeometries(WrapperTask):
+
+    def requires(self):
+        geom_types = ('iris', 'overseas_iris')
+        for geom in geom_types:
+            yield OutputAreas(geom=geom)
```
talos: This is currently broken because `OutputAreas` no longer takes a `geom` parameter.

If `OutputAreas` is going to generate all sorts of other geoms, it should be added back, even if it is ignored for now.

PR #106, opened 2016-10-13
changing logo and CARTO text

michellemho:


catalog/source/cover.tmpl
```diff
@@ -2,7 +2,7 @@
 
    Spacer 0 11cm
 
-.. image:: ../img/cartodb.png
+.. image:: ../img/logo_CARTO_positive_90.png
```
talos: This file was not committed to version control.

talos:
Addresses #19, but that shouldn't be closed til new PDF is deployed.

PR #120, opened 2016-12-09
Australia data

aspcanada:


talos:
Just did a quick review of this.  Noted a few issues:

* [ ] Linebreaks in the meta file appear to be CR only, instead of CRLF or LF, and thus aren't opening properly on the ETL box.
* [ ] Like with Brazil, the output tables appear to be separated by state, which will cause issues upon deploy.  Each final geography and data table should cover the entire of Australia.
* [ ] Source and license tags are missing. I've stubbed them in.
* [x] Geom_refs were shared between all geographies.  I've fixed this.

talos:
Great! I added another commit where I removed the abstract base classes (`BaseParams` and `BaseDataParams`).  I gave the specific reasons why in the commit f9502c1, but I'm coming to the perspective that sharing these `Params` base classes is an antipattern when writing the ETLs:

1. Tasks that don't need certain params end up getting them even though they're useless (for example, a `state` param for a `TableTask` that reads data from every single `state`).
2. Because of (1), all the params need defaults so that tasks that don't need certain params don't have to have them explicitly set.  This leads to a whole class of errors where tasks can `require` other tasks but without setting all the params necessary, and it isn't noticed because the defaults fill out the missing ones.

While there are a few duplicated params, I think having it really clear which params go with which class is way more important.  Really the only parameter I saw that could be shared between all the tasks was `year`.

talos:
I'm also having some issues with the column limit -- it looks like the output `TableTask` has more than 1600 columns.  Did you run into this at all? @aspcanada  It's preventing several of the tables from executing (I'll be able to say which ones once this completes).

talos:
Executing these now, so far so good.  Heading out for the day but will check in when I return tomorrow.

talos:
This is looking much better, I was able to run all geographies and all data, and am working on generating a catalog now.  I may have more comments once the catalog is done, as it may surface nesting or other descriptive issues.

The big bug that I found was that [we are assigning geom_ids in the Geography tasks from `ogc_fid`](https://github.com/CartoDB/bigmetadata/blob/sg-au-data/tasks/au/geo.py#L181), which is a serial number useless for joining to other tables (which is the purpose of `geom_id`).

Instead, the `geom_id` should be a SELECT from the relevant Australian census identifier for that table.  For example, in `SA1` (temp table `"au.geo".importgeography_sa1_2011_a3e9272ad2`) it should be `sa1_7digit`, which is unique across all 54805 records.  For `STE` (temp table `au.geo".importgeography_ste_2011_5ec693838b`) it should be `state_code`, which is unique across all 9 records.  The exact column name for the identifier depends on the geo resolution, so a lookup table will need to be made for that for each geom.

@aspcanada

talos:
(Take the below with a grain of salt -- this may reflect old metadata)

Having generated the catalog, I'm seeing a lack of nesting.  For example:
* [Language](http://52.71.151.140/catalog/au/language.html)
  - No nesting
  - Missing dashes between two numbers in a range.
  - Seeing a lot of trailing `Total` that means nothing, for example `Dependent children aged 13 14 years female parent Speaks other language and speaks English Total male parent Total` could be `Dependent children aged 13-14 years female parent Speaks other language and speaks English male parent`
    * columns of the above format are unclear -- the description says `MALE PARENT: Total|DEPENDENT CHILDREN AGED 13-14 YEARS`.  Where is the "female" component slipping in?
  - Seeing some `Total total`, for example `Females Total Total`.
  - Seeing redundant `Speaks other language`, and general inversion, for example `Speaks other language Arabic Females`, could just be `Females speaking Arabic`
* [Housing](http://52.71.151.140/catalog/au/housing.html)
  - Missing dashes between two numbers in a range.
  - There are some unspecified numbers, for example `0 299 Dwelling structure Flat unit or apartment`.  What is `0 299`?  The amount of rent?  The amount of combined living expenses and mortgage?
  - There is no nesting (no denominators specified).
* [Income](http://52.71.151.140/catalog/au/income.html)
  - Missing dashes between two numbers in a range.
  - A lot of unspecified numbers, for example `1500 1999 Non family households`.  Are the numbers annual gross income?  Monthly gross income?  Monthly net income?  Before or after tax?  What is the currency?
* [Age & Gender](http://52.71.151.140/catalog/au/age_gender.html)
  - Missing dashes between two numbers in a range.

Denominators are definitely specified in the metadata.  Is it possible there is a bug in the naming, and it's being ignored because of the use of `get` in tasks/au/data.py#302?

aspcanada:
Fixed issue with geo id in 23b1445

talos:
@aspcanada I re-wrote a little bit of the problematic lines so that a clear error is thrown when an expected denominator does not exist.  After that, B01 failed immediately, so I had to slightly reorder columns there (to put total_pop before male pop and female pop) in order for the denom column to exist before its numerators are created.

B02 and B01 worked after that change, I got this error:

```
  File "/usr/local/lib/python2.7/dist-packages/luigi/worker.py", line 177, in run
    new_deps = self._run_get_new_deps()
  File "/usr/local/lib/python2.7/dist-packages/luigi/worker.py", line 115, in _run_get_new_deps
    task_gen = self.task.run()
  File "/bigmetadata/tasks/util.py", line 835, in run
    for _, coltarget in self.output().iteritems():
  File "/bigmetadata/tasks/util.py", line 853, in output
    for col_key, col in self.columns().iteritems():
  File "/bigmetadata/tasks/au/data.py", line 307, in columns
    targets_dict[cols[denom_id]] = 'denominator'
KeyError: 'B34_R_0_74_Tot'
rollback Columns_BCP_B34_2011_e036229549: 'B34_R_0_74_Tot'
```

Which would suggest that B34 should be a requirement of B03, but that relationship is not defined by tasks/au/data.py#168 .

talos:
@aspcanada There are many thousands of denom failures where both the numer column and the denom column are in the same table, so I think you should write that queue functionality, as that will take care of them all.

There were only a handful where the denom was in a different table.  Take a look, these should either be errors or maybe just a missing dependency:

Setting environment variables for machine default...
colid | denom_id
--- | ---
B31_Total_PDs_Dwellings | B29_Total_dwelings
B33_Tot_Tot     | B29_Total_dwelings
B35_Total_Total | B29_Total_dwelings
B32_Total_Total | B29_Total_dwelings
B36_Total_Total | B29_Total_dwelings
B28_Tot_Tot     | B25_Total_F
B27_Tot_Families |       B25_Total_F
B26_Tot_Tot      | B25_Total_F
B34_Tot_Tot      | B29_Total_dwelings
B26_Tot_Tot      | B25_Total_F
B27_Tot_Families |         B25_Total_F
B28_Tot_Tot      | B25_Total_F
B31_Total_PDs_Dwellings | B29_Total_dwelings
B32_Total_Total | B29_Total_dwelings
B33_Tot_Tot     | B29_Total_dwelings
B34_Tot_Tot     | B29_Total_dwelings
B35_Total_Total | B29_Total_dwelings
B36_Total_Total | B29_Total_dwelings

talos:
Merged in master to grab some CI test fixes from upstream, so tests working again.

Will look in more detail this afternoon.

PR #124, opened 2017-01-31
BoundaryTags

michellemho:
- Created BoundaryTags for "interpolation_boundary" and "cartographic_boundary"
- In geometry columntasks, added appropriate tags where needed
- Incremented versions of each task

I haven't checked this locally yet, because the bigmetadata environment isn't fully set up on my laptop + I can't remember the exact obsmeta to local command to run...


talos:
@michellemho Travis is discovering legit issues with the PR now: https://travis-ci.org/CartoDB/bigmetadata/builds/213585452

Looks like a missing import?

PR #128, opened 2017-02-03
ACS Margin of Error (again)

michellemho:


ethervoid:
@michellemho what is this for? is necessary? I don't know what MoE is :)

ethervoid:
@michellemho ping?

michellemho:
@ethervoid, sorry for lack of response! MOE stands for margin of error. Every variable in the U.S. ACS survey comes with a +/- margin of error. The margin of error can be used in building predictive models for variables and assessing precision and accuracy.

It looks like I drafted code that would only create metadata structure for margin of error, but these columns remain unpopulated. The TableTask needs to be updated to populate these columns. It appears John is saying taking a look at the raw ACS tables locally is not possible, but I assume they exist somewhere in the cloud. We need to find out what the MOE columns are called in the original raw ACS tables and update the TableTask accordingly.

ethervoid:
@michellemho thank you!

Ok I'm closing this by now (I'm cleaning the number of PR we have). I've created a new issue to tackle this when possible.

PR #129, opened 2017-02-06
Canary islands fix

michellemho:
The Canary Islands geometries were missing. This is because the file path name contains an "auxiliary id" that indicates the geographical area for individual treatment of the data in various projections. 'X' for Peninsula and Balearic Islands. 'Y' for Canary Islands. Previously, we only used 'X'.


tasks/es/cnig.py
```diff
@@ -34,14 +34,15 @@ class ImportGeometry(Shp2TempTableTask):
...
 
     def requires(self):
         return DownloadGeometry(seq='114023')
 
     def input_shp(self):
         path = os.path.join('SIANE_CARTO_BASE_S_3M', 'anual', self.timestamp,
-                            'SE89_3_ADMIN_{resolution}_A_X.shp'.format(
-                                resolution=self.resolution.upper()))
+                            'SE89_3_ADMIN_{resolution}_A_{aux}.shp'.format(
+                                resolution=self.resolution.upper(),aux=self.id_aux).lower())
```
talos: Additional parameters belong on next line.

tasks/es/ine.py
```diff
@@ -84,7 +84,7 @@ def columns(self):
         cusec_id = OBSColumn(
             name=u"Secci\xf3n Censal",
             type="Text",
-            tags=[],
+            tags=[sections['spain'], subsections['boundary'],source,license],
```
talos: `GEOM_REF` should not be assigned section, subsection, source or license.

talos: This is causing duplicate entry in the catalog: https://s3.amazonaws.com/data-observatory-catalog/catalogs/canary-islands-fix/spain/boundary.html

tasks/es/cnig.py
```diff
@@ -157,16 +158,18 @@ class Geometry(TableTask):
 
     resolution = Parameter()
     timestamp = Parameter(default='20150101')
+    id_aux = Parameter()
```
talos: Instead of creating a new table with the Canary Islands, the Canary Islands should be added to the existing table.

Essentially, the `TableTask` should require two different `ImportGeometry` tasks, and then run `INSERT` once for each of them.

tasks/fr/insee.py
```diff
@@ -277,13 +277,28 @@ def columns(self):
             type='Text',
             name='DCOMIRIS',
             description='Full Code IRIS. Result of the concatenation of DEPCOM and IRIS attributes. ',
-            weight=0,
+            weight=1,
```
talos: GEOM_REF should have a weight of 0.

talos: Also, not sure why INSEE changes are leaking into the canary islands fix.  Can you remove these changes and add them to the correct branch? Looks like changes to La Caixa leaked in here, too.

michellemho: LaCaixa changes are relevant. GEOM_REF for INSEE is actually weight=0 in the france-boundary-names branch... not sure why it's 1 here and how it even got into this branch. The france-boundary-names branch will need to be merged first then.

talos: Sounds like an old one leaked in, yup let's take a look at the diff once the French work is merged!

talos:
This is failing because the filename for download needs to be uppercase except for the `.shp`.

PR #132, opened 2017-02-09
[WIP] Deployment docs

talos:


docs/source/development.rst
```diff
@@ -80,6 +80,7 @@ overwriting all output it has already created.  For example, if you have
 a :ref:`tasks.util.TempTableTask` that you've modified in the course of
 development and need to re-run:
 
+```
```
talos: This is a `.rst`, not a `.md`, so the code blocks are done a bit differently:

```
.. code:: python

    # your python code here, indented
```

docs/source/validation.rst
```diff
@@ -0,0 +1,461 @@
...
+
+.. code:: shell
+
+     make catalog
+
+1. Are there any nasty typos or missing data?
+2. Does the nesting look right?  Are there columns not nested?
+3. Are sources and licenses populated for all measures?
+4. Is a table with a boundary/timespan matrix appearing beneath each measure?
+
```
michellemho: 1. Are there any nasty typos or missing data?
     - Variable names should be unique, human-readable, and concise. If the variable needs more in-depth definition, this should go in the "description" of the variable.
2. Does the nesting look right? Are there columns not nested?
     - Variables that are denominators should also have subcolumns of direct nested variables.
     - There may be repetitive nesting if a variable is nested under two denominators, which is fine.
3. Are sources and licenses populated for all measures?
     - A ``source`` and ``license`` :ref:`tasks.util.OBSTag` must be written for new sources and licenses
4. Is a table with a boundary/timespan matrix appearing beneath each measure?
5. Are code blocks generated with a sample latitude and longitude for the area?
     - If not, hardcode the sample latitude and longitude in :ref:`tasks.meta.catalog_lonlat`

PR #138, opened 2017-02-27
Test table metawrappers

michellemho:
Adding metawrapper tasks


talos:
Testing issues on travis due to timeouts should be resolved now.

PR #140, opened 2017-03-07
BLS - QCEW File name assumption causes failure.

joofeloof:
Appreciate what you're doing here.  I've come across a small corner case on the us.bls extract.

You've implicitly assumed one would only extract full year such that q1-q4 would be contained in the input_csv file name.  Currently 2016 is named as q1-q3 so the load fails due to input_csv returned invalid file.  I don't know if when 2017.1 is available they will name it q1-q1 (seems unlikely) but I don't want to make the assumption as of yet. 

I may have overlooked something in the implementation but figured I'd at least raise it to your attention.

### DIFF
--- a/tasks/us/bls.py
+++ b/tasks/us/bls.py
@@ -34,7 +34,13 @@ class RawQCEW(CSV2TempTableTask):
         return DownloadQCEW(year=self.year)
 
     def input_csv(self):
-        return os.path.join(self.input().path, '{}.q1-q4.singlefile.csv'.format(self.year))
+        # Depending on year of extract, file name changes -- assumption that q4 is present is too strict
+        # The below may not even work when only 1 quarter of year is present? 
+        for qtr_no_ in [1, 2, 3, 4]:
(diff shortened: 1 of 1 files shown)


talos:
Thanks for the PR!

Took a quick look, my only comment would be that we should go in reverse `[4, 3, 2, 1]`, so that by default the latest quarter is chosen.  Right now, it looks like we'll always use Q1 and thus miss out on data.

Does that make sense?

joofeloof:
It does.  I figured the isfile() check would handle it either way, since we only return the filename if it is the true file contained in the archive (should only be one for a given year/singlefile download).  I haven't traced all the calls to this function though so perhaps I'm overlooking something else.

It may be overkill, but wonder if there is a better way to get the list of files in the archive itself upon extraction?  I am weary to assume this fix handles the single quarter case (we assume they name it q1-q1) but you never know with the government! :)

ethervoid:
Closing this in order to do when we have an slot for it #306

Issue #141, opened 2017-03-14
Resolve order-of-magnitude off raster summaries


talos:
The estimate of number of geometries in our raster tables are off by an order of magnitude for certain geometries, in particular in Australia.

```
gis=# select table_id, column_id, (ST_SummaryStatsagg(tile, 1, false)).sum from observatory.obs_column_table_tile_simple where column_id like 'au.%' group by
column_id, table_id;
                table_id                |  column_id   |  sum
----------------------------------------+--------------+--------
 au.geo.geography_gccsa_2011_97245b66f9 | au.geo.GCCSA |  34132
 au.geo.geography_sa3_2011_519c1855d0   | au.geo.SA3   |    249
 au.geo.geography_sos_2011_ef0a61a695   | au.geo.SOS   | 642227
 au.geo.geography_sosr_2011_4218ee06c0  | au.geo.SOSR  | 694869
 au.geo.geography_sed_2011_175a8c496d   | au.geo.SED   |    347
 au.geo.geography_poa_2011_3934ad5df1   | au.geo.POA   |  16610
 au.geo.geography_lga_2011_f929b111af   | au.geo.LGA   |  38626
 au.geo.geography_sua_2011_384b34d687   | au.geo.SUA   |  34236
 au.geo.geography_ste_2011_5ec693838b   | au.geo.STE   |  34119
 au.geo.geography_ucl_2011_1846e88735   | au.geo.UCL   |  39788
 au.geo.geography_ssc_2011_35b6f6e60f   | au.geo.SSC   |  10942
 au.geo.geography_sa1_2011_a3e9272ad2   | au.geo.SA1   |  56301
 au.geo.geography_sa2_2011_b504ddc185   | au.geo.SA2   |   2084
 au.geo.geography_sa4_2011_cf0a4addb3   | au.geo.SA4   |  34199
 au.geo.geography_ced_2011_1da5f01951   | au.geo.CED   |    694
 au.geo.geography_ra_2011_38be5b45b7    | au.geo.RA    |     21
 au.geo.geography_sla_2011_f86dbced1b   | au.geo.SLA   |   1256
```

In particular `au.geo.LGA`, which should only be ~600 for the whole of Australia.

To resolve this, several things:

* [x] Add an assertion to `generate_tile_summary` to ensure that the total of geometries in tiles generated corresponds to the number of rows in the underlying table
* [x] Fix underlying condition in that same method
* [x] Re-run tile summaries for all affected geoms
[...]

PR #142, opened 2017-03-14
Fix geom summaries

talos:
This addresses #141.

### DIFF
--- a/tasks/util.py
+++ b/tasks/util.py
@@ -415,7 +415,7 @@ def generate_tile_summary(session, table_id, column_id, tablename, colname):
                table_id=table_id)).fetchone()[0]
 
     assert abs(real_num_geoms - est_num_geoms) / real_num_geoms < 0.05, \
-            "Estimate of {} total geoms more than 25% off real {} num geoms " \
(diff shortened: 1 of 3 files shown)


tasks/util.py
```diff
@@ -248,33 +256,34 @@ def generate_tile_summary(session, table_id, column_id, tablename, colname):
         CREATE TEMPORARY TABLE raster_empty_{tablename_ns} AS
         SELECT ROW_NUMBER() OVER () AS id, rast FROM (
           WITH tilesize AS (SELECT
-            CASE WHEN ST_Area(the_geom) > 5000 THEN 250
-                 ELSE 50 END AS tilesize
-            FROM observatory.obs_table
-            WHERE id = '{table_id}'
+            CASE WHEN SUM(ST_Area({colname})) > 5000 THEN 250
```
talos: Need to check whether this causes performance degradation.

`the_geom` will drop very small isolated geoms right now.

talos: This does not cause any catastrophic performance degradation.  We can still summarize US block groups in about a minute.

PR #146, opened 2017-03-17
doc changes

mehak-sachdeva:
Just a few doc changes @talos


ethervoid:
Hi! Why did you remove the caution words about Mac? Just to know the reason :)))

talos:
I think Docker for Mac now supports sharing directories between containers and the host without mangling permissions, which was why this was commented before.

PR #149, opened 2017-03-23
Australia Area Proper Names

michellemho:
Adding proper name column for all geometries in Australia. The `proper_name` field for all geography levels must be checked because I made the assumption that all the proper names could be found as &lt;abbreviation of area\> + '_NAME' (ie. 'SA1_NAME' or 'LGA_NAME') This may not be the case for all geography levels. Source and license information also added.


michellemho:
Fix to 'SA1_NAME' to 'STATE_NAME'... Statistical Areas Level 1 do not appear to have proper names.

talos:
Checked data locally, this looks 👍 and good to merge!

PR #150, opened 2017-03-23
Faster generation of metadata

talos:
Improvements to execution speed for TableTask by more quickly obtaining output from tags.


javitonino:
Ugh, could you merge master into this so the diffs are more legible. It should also fix the conflicts and the tests.

tasks/util.py
```diff
@@ -972,7 +983,17 @@ def complete(self):
...
+            cnt = current_session().execute(
+                '''
+                SELECT COUNT(*)
+                FROM observatory.obs_tag
+                WHERE id IN ('{ids}') AND version = '{version}'
+                '''.format(
+                    ids="', '".join(self.tagids.values()),
+                    version=self.version()
+                )).fetchone()[0]
+            return cnt == len(self.tagids.values())
```
javitonino: We can probably do a `TagsTarget` that wraps multiple tags checks, and use that instead of patching `complete`. Anyway, probably something to do with the models refactor.

javitonino:
LGTM, although it adds to the `complete()` of the TagsTask in order to get speed. @ethervoid can you take a look? This seems awfully related to your parallelization struggles, maybe you have more insights.

tasks/util.py
```diff
@@ -972,7 +983,17 @@ def complete(self):
         if deps and not all([d.complete() for d in deps]):
             return False
         else:
-            return super(TagsTask, self).complete()
+            # bulk check that all tags exist at proper version
+            cnt = current_session().execute(
+                '''
+                SELECT COUNT(*)
+                FROM observatory.obs_tag
+                WHERE id IN ('{ids}') AND version = '{version}'
```
ethervoid: version is going to be 0. in output it takes from the tag model object

tasks/util.py
ethervoid: From my point of view, we should add this check in the `exists` method of the TagTarget which is where it belongs. Now it does some checks for versions and so on over the model. Maybe we should change it to a plain SQL. 

I know it'll be like 3-4 queries instead of one but we should go with the flow of Luigi and avoid making this kind of things in `complete`

antoniocarlon: So, do you agree that we can leave it like this and add a `TO DO` task to the refactor to fix it discussing the implementation details?

javitonino: Pretty much the whole point of this PR is to be able to do this (count tags with a single SQL call) for speed. But it makes readability/maintainability poorer, and we intend to review more of the metadata code in the future anyway.

So let's make a decision, do we want to merge this, or trash it? I'm in for trashing it unless this is solving a real problem. So questions:
- Anyone has come into a situation where tag generation was slow enough? I haven't.
- Would this help with the parallelization problems? I recall there were problems checking for tags existence regarding this. Do you think this would solve it, @ethervoid?

antoniocarlon: As we want to do a refactor and we are not currently finding slow tag generation issues I would trash this. Also, @ethervoid is not happy adding this kind of logic to the `complete` and he is our Luigi expert so we should discuss a new approach when we do the refactor.

ethervoid: I'll trash it too. This doesn't help with the parallelization problem and if we don't have Tag generation complains 🔥

javitonino:
To the 🗑! A bit of a pity after taking the time to bring it up to speed with the rest of the code. Hope we can get some time for a bigger refactor of this soon :)

PR #151, opened 2017-03-23
add tags to thaipop.name

michellemho:
Tags missing from name


talos:
👍 just added a commit to bump the version number for the table.

PR #152, opened 2017-03-23
Canada geometry names

michellemho:
Adding geometry names for Canadian provinces + territories, census tracts, census divisions, census subdivisions, and census metropolitan areas and census agglomerations.

Assuming that all name codes are "RESOLUTION+NAME" (eg. "CTNAME")

### DIFF
--- a/tasks/ca/statcan/geo.py
+++ b/tasks/ca/statcan/geo.py
@@ -140,6 +140,7 @@ class GeographyColumns(ColumnsTask):
         geom_name = OBSColumn(
             type='Text',
             weight=1,
+            name='Name of ' + GEOGRAPHY_NAMES[self.resolution],
             targets={geom: GEOM_NAME},
             tags=[sections['ca'], subsections['names']]
         )
--- a/tasks/ca/statcan/geo.py
+++ b/tasks/ca/statcan/geo.py
@@ -175,7 +175,7 @@ class Geography(TableTask):
     def populate(self):
         session = current_session()
         session.execute('INSERT INTO {output} '
-                        'SELECT {name} as geom_name '
+                        'SELECT {name} as geom_name, '
                         '       {code}uid as geom_id, '
                         '       wkb_geometry as geom '
                         'FROM {input} '.format(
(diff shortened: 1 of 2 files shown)


talos:
Looks like TIGER changes slipped in here.  Will have to hold off on review until #153 changes are merged.

ethervoid:
Is this ok? after fixing the conflic is ok to merge?

michellemho:
I believe so @ethervoid.

ethervoid:
To clarify for a newbie hehe, what is the purpose of including this names?

michellemho:
A while back, John and I added in a "Proper Names" subsection tag so we could log the proper names of geometries. Most of the lower level geographies (for example Census Tracts in the US) do not have proper names. But higher level geographies such as Provinces or States will have proper names. We wanted to capture those.

michellemho:
The process was to look at the shapefile or original raw data of the downloaded shapes. Usually, that data comes with a proper name, as well as a "geo_id" column which we are already adding.

All I needed to do was see what the proper name column was labeled in the original raw data. In the case of Canada, it appears the proper names in the geography tables were labeled 'CTNAME', 'PRNAME', 'CDNAME', 'CSDNAME', and 'CMANAME'. I would refer to these labels in the `populate` method of the TableTask.

I would also add proper names to the Geography Columns task, giving it a target type of  `targets={geom: GEOM_NAME}`.

Sometimes, the proper name is exactly the same as the geo_id (usually the case for smaller geographies). I still added them; even if it means duplicate information.

tasks/ca/statcan/geo.py
```diff
@@ -160,9 +177,11 @@ def columns(self):
     def populate(self):
         session = current_session()
         session.execute('INSERT INTO {output} '
-                        'SELECT {code}uid as geom_id, '
+                        'SELECT {name} as geom_name, '
+                        '       {code}uid as geom_id, '
                         '       wkb_geometry as geom '
                         'FROM {input} '.format(
+                            name=GEOGRAPHY_PROPERNAMES[self.resolution()],
```
antoniocarlon: This line should be `name=GEOGRAPHY_PROPERNAMES[self.resolution],` (removing the parentheses)

PR #154, opened 2017-03-24
Mexico shape NOMBRE added to certain resolutions

michellemho:
NOMBRE column added for entidad, localidad_urbana_y_rural_amanzanadas, municipio, and servicios_area.

### DIFF
--- a/tasks/mx/inegi.py
+++ b/tasks/mx/inegi.py
@@ -15,10 +15,15 @@ RESOLUTIONS = (
     'municipio', 'servicios_area',
 )
 
-RESNAMES = (
-    'ageb', 'entidad', 'localidad_urbana_y_rural_amanzanada', 'manzana',
-    'municipio', 'servicios_area',
-)
+RESNAMES = {
+    'ageb': 'Census area (urban areas only)',
+    'entidad': 'State',
+    'localidad_rural_no_amanzanada': 'Localidades (rural)',
+    'localidad_urbana_y_rural_amanzanada': 'Localidades (urban)',
+    'manzana': 'Census block (urban areas only)',
+    'municipio': 'Municipios',
+    'servicios_area': 'Service areas',
+}
 
 RESPROPNAMES = (
     'entidad', 'localidad_urbana_y_rural_amanzanada',
--- a/tasks/mx/inegi.py
+++ b/tasks/mx/inegi.py
@@ -268,7 +268,7 @@ class Geography(TableTask):
         output_cols = ', '.join(column_targets.keys())
         input_cols = ','.join(['{}::{}'.format(colname, ct.get(session).type)
                                for colname, ct in column_targets.iteritems()])
-        session.execute('INSERT INTO {output} {output_cols} '
+        session.execute('INSERT INTO {output} ({output_cols}) '
                         'SELECT {input_cols} '
                         'FROM {input} '.format(
                             output=self.output().table,
--- a/tasks/mx/inegi.py
+++ b/tasks/mx/inegi.py
@@ -203,7 +203,7 @@ class GeographyColumns(ColumnsTask):
         }
 
     def version(self):
-        return 9
+        return 10
 
     def columns(self):
         input_ = self.input()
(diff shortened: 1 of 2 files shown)


tasks/mx/inegi.py
```diff
@@ -216,19 +221,22 @@ def columns(self):
...
-            type='Text',
-            weight=0,
-        )
+        cols = OrderedDict([('the_geom', geom),
+                            ('cvegeo', geom_ref)
+                        ])
+
+        if self.resolution in RESPROPNAMES:
+            cols['nomgeo'] = OBSColumn(
+                type='Text',
```
talos: This will result in overlapping IDs, because `id` is not specified separately for each possible source geom.

tasks/mx/inegi.py
```diff
@@ -253,12 +261,17 @@ def columns(self):
 
     def populate(self):
         session = current_session()
-        session.execute('INSERT INTO {output} (the_geom, cvegeo) '
-                        'SELECT wkb_geometry, cvegeo '
+        column_targets = self.columns()
+        output_cols = ', '.join(column_targets.keys())
+        input_cols = ','.join(['{}::{}'.format(colname, ct.get(session).type)
+                               for colname, ct in column_targets.iteritems()])
+        session.execute('INSERT INTO {output} {output_cols} '
```
talos: `output_cols` will need to be in parentheses.

tasks/mx/inegi.py
```diff
@@ -253,12 +262,17 @@ def columns(self):
...
     def populate(self):
         session = current_session()
-        session.execute('INSERT INTO {output} (the_geom, cvegeo) '
-                        'SELECT wkb_geometry, cvegeo '
+        column_targets = self.columns()
+        output_cols = ', '.join(column_targets.keys())
+        input_cols = ','.join(['{}::{}'.format(colname, ct.get(session).type)
+                               for colname, ct in column_targets.iteritems()])
+        session.execute('INSERT INTO {output} ({output_cols}) '
+                        'SELECT {input_cols} '
```
talos: `{input_cols}` here is actually pulling from the same definition as output columns -- so this doesn't work.  For example, on the output, the geometry is called `the_geom`, but on input it's called `wkb_geometry`.  This needs to be adjusted to match column names from the input table.

talos: @michellemho could you take a look at this?

tasks/mx/inegi.py
```diff
@@ -216,19 +221,23 @@ def columns(self):
...
+        cols = OrderedDict([('the_geom', geom),
+                            ('cvegeo', geom_ref)
+                        ])
+
+        if self.resolution in RESPROPNAMES:
+            cols['nomgeo'] = OBSColumn(
+                id=self.resolution + '_name',
+                type='Text',
+                weight=1,
+                name= 'Nombre de {}'.format(self.resolution),
```
talos: We're using English for our Mexican measures, so we should change this to "Name of".

talos:
@michellemho could you take a look at the comments for fixes please?

ethervoid:
What about this? What is its state?

javitonino:
Works fine, but we have encoding problems

```sql
WITH data AS (
  SELECT 1 as id,
  ST_Buffer(ST_SetSRID(ST_Point(-99.17, 19.413), 4326), 0.01) the_geom
),
meta AS (
  SELECT cdb_observatory.OBS_GetMeta(ST_SetSRID(ST_Extent(the_geom), 4326),
  '[{"numer_id": "mx.inegi.municipio_name",
     "normalization": "areaNormalized"}]'
  , 1, 1) meta
FROM data)
SELECT id, (data->0->>'value')::Text as per_sq_km
FROM cdb_observatory.OBS_GetData(
       (SELECT ARRAY_AGG((the_geom, id)::geomval) FROM data),
       (SELECT meta FROM meta));
```
returns `CuauhtÃ©moc`

Or, jsut do `select nomgeo from observatory.obs_366a1365508f0a27c925858ba2afb2d09884e576;`

PR #156, opened 2017-04-04
Spain geo names

michellemho:
Added a new GeoNameColumns task to add proper names (aka "rotulos" or "labels") to CNIG shapes. Noted that CNIG has a new descargas (download) portal which makes the URL for the Download task no longer work.


tasks/es/cnig.py
```diff
@@ -152,6 +156,31 @@ def columns(self):
...
+                'subsections':SubsectionTags(),
+                'sections':SectionTags(),
+            }
+
+    def columns(self):
+        cols = OrderedDict()
+        session = current_session()
+        for colname, coltarget in self.input()['geom_cols'].iteritems():
+            cols['name_' + colname] = OBSColumn(
+                type='Text',
```
talos: This is missing a unique ID.

tasks/es/cnig.py
```diff
@@ -152,6 +156,31 @@ def columns(self):
...
+                name='Proper name of {}'.format(colname),
+                tags=[self.input()['subsections']['names'],self.input()['sections']['spain']],
+                weight=1,
+                targets={coltarget: GEOM_NAME},
+            )
+        return cols
+
 
 class Geometry(TableTask):
```
talos: Version number should be incremented.

tasks/es/cnig.py
```diff
@@ -191,7 +222,7 @@ def geom_ref_colname(self):
     def populate(self):
         session = current_session()
         query = 'INSERT INTO {output} ' \
-                'SELECT {geom_ref_colname} geom_ref, wkb_geometry the_geom ' \
+                'SELECT {geom_ref_colname} geom_ref, rotulo, wkb_geometry the_geom ' \
```
talos: It's always `rotulo`?

tasks/es/cnig.py
michellemho: is that a problem for the GeomRefColumns task as well? I copied the code for that task over to this one, there weren't unique IDs for the GeomRefs columns there either...?

tasks/es/cnig.py
michellemho: yes, I believe I checked that.

PR #157, opened 2017-04-05
Brazil geonames

michellemho:
Adding proper names for all Brazil geometries.


talos:
@michellemho It looks like there are some undefined names leaking through?  Take a look at the sample Catalog: https://s3.amazonaws.com/data-observatory-catalog/catalogs/brazil-geonames/br/names.html

tasks/br/geo.py
```diff
@@ -200,18 +214,38 @@ def columns(self):
...
+                id = self.resolution + '_name',
+                name="Name of the {}".format(GEOGRAPHY_NAMES[self.resolution]),
+                type='Text',
+                weight=1,
+                tags=[sections['br'], subsections['names']],
+                targets={geom: GEOM_NAME}
+            )
+            cols['{}'.format(GEOGRAPHY_PROPERNAMES[self.resolution])] = geom_name
+        else:
+            for region in GEOGRAPHY_PROPERNAMES[GEO_I]:
```
talos: What does this do?  It looks like for census tract geographies, it adds one column per region, but I don't understand why.

michellemho: Census tracts in Brazil (aka setores_censitarios) don't have proper names, but they are identified by a mesoregion and microregion. So this columnstask adds a column for each region.

talos: Makes sense.  Could you just add a little comment in the code so it's clearer that's what's going on here?

PR #159, opened 2017-04-10
add slack build notifications from travis, including location of buil…

talos:
…d catalog


ethervoid:
We're not going to do this by now

Issue #169, opened 2017-06-14
Reduce number of variables


michellemho:
I think there are too many variables in the Data Observatory. This makes searching, organizing, and navigating the measurements nearly impossible with the systems we have now (the Catalog and the Builder UI). I think we should consider reducing the number of variables to just few hundred key variables per country to ensure quality before adding more quantity.

There are thousands of detailed variables that are unnecessary. Brazil and Australia are especially bad. For example there is a variable in Brazil for "Daughters of 38 year-old Guardian and Spouse" (br.data.Pessoa07_V039) ... this is too much detail. There are thousands more that are similar. Just look how massive this page is! http://cartodb.github.io/bigmetadata/br/age_gender.html

michellemho:
This issue should be fixed before tackling the names and descriptions:
https://github.com/CartoDB/bigmetadata/issues/168

michellemho:
I'll need more feedback on this, but I believe users are most interested in getting total population, population breakdowns by age and gender, median income, and maybe just a handful of other variables.

John hand-curated the variables available for the United States from the American Community Survey (ACS). There are thousands of variables to choose from-- he only included a few hundred. These names and descriptions were manually written into the ETL process in the ColumnsTask of the acs.py file.

juanignaciosl:
Yep, we can probably begin with a set of generic values that are available for most of the countries and let the most specific ones aside. In the future, we might be interested in providing users the possibility to add them to their accounts only if they're interested in them.

michellemho:
This is my first pass to come up with bare minimum variables + key variables for the Data Observatory. When we start adding in censuses, we should prioritize aligning as close as possible to these variables first. I chose not to included a few subsections that already exist such as transportation, migration, and society because I don't think any of those are "key" variables. If other variables exist and are available, they should be added on an ad-hoc basis by country.

## Minimum demographic variables for all countries:
 
**Age & Gender**

- Total Population
- Male Population
- Female Population
- Population by age groups (varies country to country)
- Population by age groups and gender (varies country to country)

**Households (Families)**

- Total households

**Housing**
- Total housing units
 
## Key demographic variables (availability varies by country)
 
**Households (Families)**

- Average household size
- Number of households by size
- Number of people by marriage status (single, married, divorced, separated, widowed)
- Number of households or families with children

 
**Housing**

- Occupied housing
- Owner-occupied housing
- Renter-occupied housing
- Vacant housing
[...]

PR #189, opened 2017-06-28
Public tags

michellemho:
Creating and adding a new tag called "public" to 109 variables from Canada's Census and NHS survey. This PR is to start reducing variables to a few hundred key variables as discussed in this issue: https://github.com/CartoDB/bigmetadata/issues/169.

After running 

`make -- run ca.statcan.cols_census CensusColumns`
and `make -- run ca.statcan.cols_nhs NHSColumns`

You can find the 109 Canadian variables tagged as "public" by opening psql (`make psql`) and querying:

```
select C.* from observatory.obs_column C inner join observatory.obs_column_tag T on C.id = T.column_id where T.column_id like 'ca.%' and T.tag_id like 'tags.public'
```

This tag can be used to isolate the limited key variables we can publish to Builder users via the DO UI.

### DIFF
--- a/tasks/ca/statcan/cols_census.py
+++ b/tasks/ca/statcan/cols_census.py
@@ -9752,7 +9752,7 @@ class CensusColumns(ColumnsTask):
             type='Numeric',
             weight=3,
             aggregate='sum',
-            tags=[ca, unit_households, subsections['families']],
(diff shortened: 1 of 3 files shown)


ethervoid:
Looks good to me, merging!

PR #194, opened 2017-07-11
Boundarytags (Interpolation Boundary tag for US Congressional Districts)

michellemho:
Adding US Congressional Districts for interpolation boundary tag. There is no evidence found that these boundaries are not full coverage of the US.

Preview of geography file:
https://observatory.carto.com/tables/obs_664d0ac6cc93a8f8d83553cb454738843132cb4c/public/map


ethervoid:
Added in #196

PR #196, opened 2017-07-18
Sort geometries

ethervoid:
### Additions

- Added congressional districts and zcta5 as interpolate geometries
- Changed the weight for the TIGER geometries in order to make it possible to be sorted

### Fixes

- Added version method for the Survey task in Canadian NHS data
- Removed unused parameter in us.tiger.ShorelineClip task which makes the geometries generation fail

### DIFF
--- a/tasks/us/census/tiger.py
+++ b/tasks/us/census/tiger.py
@@ -666,10 +666,10 @@ class ShorelineClip(TableTask):
         session = current_session()
         stmt = ('INSERT INTO {output} '
                 'SELECT geoid, ST_Union(ST_MakePolygon(ST_ExteriorRing(the_geom))) AS the_geom, '
-                '       MAX(aland) AS aland, cdb_observatory.FIRST(name) AS name '
+                '       MAX(aland) AS aland '
                 'FROM ( '
                 '    SELECT geoid, (ST_Dump(the_geom)).geom AS the_geom, '
-                '           aland, name '
+                '           aland '
                 '    FROM {input} '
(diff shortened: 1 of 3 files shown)


tasks/us/census/tiger.py
```diff
@@ -664,10 +666,10 @@ def populate(self):
         session = current_session()
         stmt = ('INSERT INTO {output} '
                 'SELECT geoid, ST_Union(ST_MakePolygon(ST_ExteriorRing(the_geom))) AS the_geom, '
-                '       MAX(aland) AS aland, cdb_observatory.FIRST(name) AS name '
+                '       MAX(aland) AS aland '
```
juanignaciosl: I remember speaking about a _missing_ `name` column. Could you give me some context about why is this one not needed, please?

ethervoid: I reviewed the history of that file, and what changes John and Michelle made. Looks like they went back and forth adding name and removing it and for what I understood looks like this one was a missing part of that changes that are not needed anymore. You could check it looking at the history of the tiger.py class

juanignaciosl: 👍  thanks!

michellemho: I'm not exactly sure about this particular part of the code... does this have to do with the adding of the proper name for certain geometries?

ethervoid: It looks like yes. Following the stream of changes between John and you I came to the conclusion, hopefully not wrong, that this is some kind of remain from that changes.

I wasn't able to check that property anywhere and was making the task fail. Also, checking the production tables we don't have a name column.

PR #164, opened 2017-06-06
Bring nested catalog to parity with previous catalog

talos:
**New**
* Takes advantage of nesting from Sphinx
* Tabbed SQL selection for code snippets
* Much improved code snippets
* Indication of number of subcolumns
* Simpler sample latlng (fixed lonlat/latlng confusion in function name)
* Removed catalog image generation code

**Fixes**
- Recursive generation (supports 3+ levels). Fixes #239 Includes significant refactor (extracting functionality into smaller methods)
- Style adjustments (tab width overflow and bigger headers)
- Requires running https://gist.github.com/javitonino/c83c9229d46ee8ade71c17804f1a1a98 in old instance to adjust column types (due to https://github.com/CartoDB/bigmetadata/commit/31318a3c81585fa8f9c6c9a0d76449f48d5b070b)

PR #237, opened 2017-08-09
Dont recreate metadata with catalog

javitonino:
Extracted from https://github.com/CartoDB/bigmetadata/pull/164

Only the part regarding not rerunning OBSMetaToLocal on every catalog generation.


tasks/carto.py
```diff
@@ -1412,8 +1420,11 @@ def run(self):
             session.commit()
             raise
 
-    def complete(self):
-        return getattr(self, '_complete', False)
+    def output(self):
+        tables = ['obs_meta', 'obs_meta_numer', 'obs_meta_denom',
```
javitonino: Good catch! I forgot about that one.

PR #238, opened 2017-08-09
Solving Canadian census import

### LINKED ISSUE #195: Canada NHS CMA data has change or the task is not working
Re-executing the download of the Canadian NHS CMA data, I've found two issues with this dataset that make the task fail.

### File with "wrong" headers

Now there is a file called  `99-004-XWE2011001-201-DQ` which has the following content:
```csv
"Census Metropolitan Areas (CMAs) and Census Agglomerations (CAs) - Geographies which exclude National Household Survey data for one or more incompletely enumerated Indian reserves or Indian settlements"
"Geo_Code","Geo_Name","Geo_Type"
"421","Québec","CMA"
"462","Montréal","CMA"
"590","Sault Ste. Marie","CA"
"602","Winnipeg","CMA"
"935","Victoria","CMA"
```

`make -- run ca.statcan.data AllNHSTopics`

When the ETL tries to process that file, fails [here](https://github.com/CartoDB/bigmetadata/blob/master/tasks/ca/statcan/util.py#L262)

The error is: `ValueError: Invalid file: tmp/ca.statcan.data/DownloadData_cma_nhs_5600a4c4b1/99-004-XWE2011001-201-DQ.CSV^M`

### Data file with multiples parts for the same geocode
The data file called `99-004-XWE2011001-201` has some geocodes with what rows with the total data and parts for different zones, for example quebec and New Brunswick). Here a sample:
```csv
330,New Brunswick,Campbellton,CA,26.1,Income of individuals in 2010,  65 years and over (%),,30.1,,28.2,,31.5,
330,New Brunswick,Campbellton (New Brunswick part),CA,26.3,Income of individuals in 2010,  65 years and over (%),,30.1,,28.2,,31.5,
330,New Brunswick,Campbellton (Quebec part),CA,25.6,Citizenship,Total population in private households by citizenship,1,3165,,1500,,1665,
```

This makes that the `SplitAndTranspose` task generates a rows for every Topic with that geocode than includes the three parts and it fails to make the COPY to the database because it exceeds the number of columns.

The error is:
```
ERROR: ERROR:  extra data after last expected column^M
CONTEXT:  COPY copydatatotable_cma_census_t009_c52ae25de4, line 18: "330,15270,7215,8055,2595,1210,1380,12675,6005,6675,8325,4180,4145,1205,605,605,7115,3575,3540,6200,3..."^M
```

// @michellemho for awareness

antoniocarlon:
Solving Canadian census import (Indian reserves) and multiple parts for the same zone

fixes #195

### DIFF
--- a/tasks/ca/statcan/data.py
+++ b/tasks/ca/statcan/data.py
@@ -53,20 +53,13 @@ class DownloadData(BaseParams, DownloadUnzipTask):
 
 
 class SplitAndTransposeData(BaseParams, Task):
-    INDIAN_RESERVES_SUFFIX = '-DQ'
-    DIVISION_COLUMN_NAMES = {
+    DATA_QUALITY_SUFFIX = '-DQ'
+    DIVISION_SPLITTED = {
         GEO_CT: None,
         GEO_PR: None,
         GEO_CD: None,
         GEO_CSD: None,
-        GEO_CMA: 'cma_ca_name'
-    }
-    DIVISION_SUFFIX = {
-        GEO_CT: None,
-        GEO_PR: None,
-        GEO_CD: None,
-        GEO_CSD: None,
-        GEO_CMA: 'part)'
+        GEO_CMA: ('cma_ca_name', r'part\)$')
     }
 
     def requires(self):
@@ -80,12 +73,17 @@ class SplitAndTransposeData(BaseParams, Task):
         ))
         in_csv_files = []
         for in_csv_file in infiles.strip().split('\n'):
-            if not os.path.splitext(in_csv_file)[0].endswith(self.INDIAN_RESERVES_SUFFIX):
+            if not self._is_data_quality_file(in_csv_file):
                 in_csv_files.append(in_csv_file)
             else:
                 LOGGER.warning('Ignoring file %s' % in_csv_file)
         os.makedirs(self.output().path)
-        StatCanParser(self.DIVISION_COLUMN_NAMES[self.resolution], self.DIVISION_SUFFIX[self.resolution]).parse_csv_to_files(in_csv_files, self.output().path)
+        StatCanParser(self.DIVISION_SPLITTED[self.resolution]).parse_csv_to_files(in_csv_files, self.output().path)
+
+    def _is_data_quality_file(self, csv_file):
+        if os.path.splitext(csv_file)[0].endswith(self.DATA_QUALITY_SUFFIX):
+            return True
+        return False
 
     def output(self):
         return LocalTarget(os.path.join('tmp', classpath(self), self.task_id))
--- a/tasks/ca/statcan/data.py
+++ b/tasks/ca/statcan/data.py
@@ -55,11 +55,20 @@ class DownloadData(BaseParams, DownloadUnzipTask):
 class SplitAndTransposeData(BaseParams, Task):
     IGNORED_FILE_SUFFIXES = ('-DQ',)
     DIVISION_SPLITTED = {
-        GEO_CT: None,
-        GEO_PR: None,
-        GEO_CD: None,
-        GEO_CSD: None,
-        GEO_CMA: ('cma_ca_name', (r'part\)$',))
+        SURVEY_CEN: {
+            GEO_CT: None,
+            GEO_PR: None,
+            GEO_CD: None,
+            GEO_CSD: None,
+            GEO_CMA: ('cmaca_name', (r'part\)$',))
+        },
+        SURVEY_NHS: {
+            GEO_CT: None,
+            GEO_PR: None,
+            GEO_CD: None,
+            GEO_CSD: None,
+            GEO_CMA: ('cma_ca_name', (r'part\)$',))
+        }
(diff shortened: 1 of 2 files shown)


tasks/ca/statcan/data.py
```diff
@@ -62,9 +78,14 @@ def run(self):
             survey_code=SURVEY_CODES[self.survey],
             geo_code=GEOGRAPHY_CODES[self.resolution]
         ))
-        in_csv_files = infiles.strip().split('\n')
+        in_csv_files = []
```
ethervoid: I'll extract this to a new method

tasks/ca/statcan/data.py
```diff
@@ -53,6 +53,22 @@ def download(self):
 
 
 class SplitAndTransposeData(BaseParams, Task):
+    INDIAN_RESERVES_SUFFIX = '-DQ'
+    DIVISION_COLUMN_NAMES = {
```
ethervoid: I have mixed feelings about this approach. I like the idea but:

* Why do include the division structure inside the `SplitAndTranspose` class instead of inside the `StatCanParser`
* I think that maybe is better to have a tuple (column, prefix) and prefix should be a regexp instead a literal
* Maybe we could make a new class and/or at least one method to check this scenerio

What do you think?

tasks/ca/statcan/util.py
```diff
@@ -259,17 +267,23 @@ def parse_csv_to_files(self, csv_paths, output_dir):
                     # Need to handle csv files where header row isn't row #1
                     header_row_index, self._header = self.get_header_row(csvfile)
                     if header_row_index == -1:
-                        raise ValueError('Invalid file')
+                        raise ValueError('Invalid file', csv_path)
```
ethervoid: Nice :)

tasks/ca/statcan/data.py
```diff
@@ -53,6 +53,22 @@ def download(self):
 
 
 class SplitAndTransposeData(BaseParams, Task):
+    INDIAN_RESERVES_SUFFIX = '-DQ'
```
ethervoid: Nice to know what that suffix means. Where did you get it? I was looking for it and I was unable!

antoniocarlon: Here you can find the data quality information: http://www12.statcan.gc.ca/census-recensement/2016/dp-pd/prof/details/Page.cfm?Lang=E&Geo1=PR&Code1=62&Geo2=&Code2=&Data=Count&SearchText=Nunavut&SearchType=Begins&SearchPR=01&B1=All&GeoLevel=PR&GeoCode=62#dqf1-popup

Maybe it would be better to change INDIAN_RESERVES_SUFFIX to DATA_QUALITY_SUFFIX

tasks/ca/statcan/data.py
antoniocarlon: - The division structure is inside the `SplitAndTranspose` to decouple the division from the `StatCanParser` as it only has to manage (parse, transpose) `CSV` files
- A tuple would be great but it's just another way of organizing the code. The regexp may be overkilling.
- I'm not sure if I get you... What would this new method do?

ethervoid: * But `StatCanParser` know what to do with columns (for example what columns use, or find or...) maybe this means to move that part too
* Use of a tuple or an object or a merged data structure is to improve readability because two different dictionaries that are related is better to have them merged.
* Regexp would be great to deal with future problems like this. My suggestion is the use regexp in general for this kind of scenarios instead of literal substring so in the future, we are able to execute the regexp which from my point of view is more flexible. I have really bad experiences using substrings (if the Canada census adds a space between the `t` and the parentheses we are in trouble again).
* I'm talking about the line 283-286, extract a method to check for that condition to improve readability

tasks/ca/statcan/data.py
ethervoid: yes I like more `DATA_QUALITY_SUFFIX` too

tasks/ca/statcan/data.py
javitonino: Pitching in:
1. and 2. 
I am torn here. In one hand, I like Antonio's point that the parser should just parse a file, and receive a configuration. On the other hand, it is true that the parser already has a lot of configuration, albeit, hardcoded, and I don't like passing two separate parameters as two variables (specially since they are very closely related).

3. I am not a fan of overenginering, simplest solution is best and so on. But changing the string by a regex is quite simple and gives a lot of flexibility. I like it.
4. Agreed. It would be nice to have a method do the check if the row should be skipped in order to keep the same level of abstraction in the loop.

ethervoid:
Added some minor suggestions to make it easier to add changes in the future

ethervoid:
With the last changes ❤️  :shipit:

javitonino:
Acceptance notes:
- I had to bump version (see last commit)
- I also had to delete the temporary tables `DROP SCHEMA "ca.statcan.data" CASCADE`

Otherwise, everything seem to have imported correctly.

Issue #245, opened 2017-08-14
Modify the weights for 


antoniocarlon:
Change the weight for each resolution on the [tasks/ca/statcan/geo.py](https://github.com/CartoDB/bigmetadata/blob/master/tasks/ca/statcan/geo.py#L96-L102).

Visually inspecting the geometries it seems that as PR > CD > CMA > CSD > CT, he correct weights must be

```
    weights = {
        GEO_PR: 1,
        GEO_CD: 2,
        GEO_CMA: 3,
        GEO_CSD: 4,
        GEO_CT: 5,
    }
```

PR #246, opened 2017-08-14
Changed canadian resolution weights

antoniocarlon:
#245 Changed canadian resolution weights according to the feature size for each resolution

### DIFF
--- a/tasks/ca/statcan/geo.py
+++ b/tasks/ca/statcan/geo.py
@@ -102,7 +102,7 @@ class GeographyColumns(ColumnsTask):
     }
 
     def version(self):
-        return 11
+        return 12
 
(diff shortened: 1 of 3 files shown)


tasks/ca/statcan/geo.py
```diff
@@ -94,11 +94,11 @@ class GeographyColumns(ColumnsTask):
     resolution = Parameter(default=GEO_PR)
 
     weights = {
+        GEO_PR: 1,
```
javitonino: We are missing a version bump for this task.

antoniocarlon:
To force the execution of `GeographyColumns` when executing `AllGeographies`, we need to bump the version of the `Geography` task too

Issue #244, opened 2017-08-14
Incorrect tile generation (percent fill and geometry count)


antoniocarlon:
The `generate_tile_summary` ([tasks/util.py](https://github.com/CartoDB/bigmetadata/blob/master/tasks/util.py#L342-L348)) computes incorrect values for `percent_fill` (and maybe also for `cnt`).

There is a smell on the `WHEN geom @ vector.the_geom THEN 1` as it only takes into account the boundary of the geometries to tell whether the pixel area is fully contained within a geometry.

Also, it may be good to take a look at the method used to perform clips/intersections (`st_clip = 'ST_ClipByBox2D'`). As[ the documentation ](https://postgis.net/docs/ST_ClipByBox2D.html)states that it "Clips a geometry by a 2D box in a fast but possibly dirty way".

PR #248, opened 2017-08-17
244 tile summary generation review

antoniocarlon:
#244

Solves a problem calculating the percent fill value for each pixel. The algorithm was miscalculating some tiles with zero or very low coverage (see the green pixels on the sea):

201708161716_before

Changing the spatial relation operator from `@` to `ST_Within` solves most cases:

201708161716_after1

Regenerate the band 2 setting 1 as the minimum value to include very small geometries that would be considered spurious due to rounding (thanks to @javitonino for the idea):

201708161716_after2

Reduce the tolerance for the simplification from 0.0005 (about 50 meters) to 0.00001 (about 1 meter) to improve the percent fill calculation.

Using `ST_ClipByBox2D` instead of `ST_Intersection` doesn't seem to have any negative effect (some geometries from several resolutions have been picked for testing). The algorithm shows a performance boost when using `ST_ClipByBox2D` as expected. 

~~As the documentation for `ST_ClipByBox2D` states that _The output geometry is not guaranteed to be valid_, I have added `ST_MakeValid` to its result to increase robustness.~~ **Note:** `ST_MakeValid` checks validity using `ST_IsValid` but it can be (very) costly depending on the size (number of vertices) of the geometries.

### DIFF
--- a/tasks/util.py
+++ b/tasks/util.py
@@ -46,6 +46,7 @@ def get_logger(name):
     logger.addHandler(handler)
     return logger
 
+
 LOGGER = get_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


ethervoid:
LGTM :shipit:

javitonino:
Tested this. It works, but it takes around double the time to generate geometry rasters. This adds 5-10 minutes for big geometries (provinces), with little effect (1-2 minutes) in smaller geometries.

I think it is a good trade-off, since the old code was fast, but totally wrong 😅

PR #254, opened 2017-08-25
Solving BR data downloading issues

antoniocarlon:
#252

Ignoring files matching the state in lowercase to avoid mathing files like *1_Documentacao_Agregado_dos_Setor**es**_2010_20150527.zip* when checking for **ES**. Taking into account that some states can be splitted into different files

### DIFF
--- a/tasks/br/data.py
+++ b/tasks/br/data.py
@@ -59,7 +59,6 @@ class LicenseTags(TagsTask):
         ]
 
 
-
 class DownloadData(DownloadUnzipTask):
 
     state = Parameter()
@@ -70,15 +69,17 @@ class DownloadData(DownloadUnzipTask):
         cmd += ' | '
         cmd += 'awk \'{print $9}\''
         cmd += ' | '
-        cmd += 'grep -i {state}_[0-9].*zip$'.format(state=self.state)
+        cmd += 'grep {state}_[A-Za-z_]*[0-9].*zip$'.format(state=self.state.upper())
 
         return shell(cmd)
 
     def download(self):
-        filename = self._get_filename()
-        shell('wget -O {output}.zip {url}{filename}'.format(
-            output=self.output().path, url=self.URL, filename=filename
-        ))
+        filenames = self._get_filename().splitlines()
+
+        for filename in filenames:
+            shell('wget -O {output}.zip {url}{filename}'.format(
+                output=self.output().path, url=self.URL, filename=filename
+            ))
 
 
 class ImportData(CSV2TempTableTask):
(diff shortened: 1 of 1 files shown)


tasks/br/data.py
```diff
@@ -70,15 +69,17 @@ def _get_filename(self):
...
     def download(self):
-        filename = self._get_filename()
-        shell('wget -O {output}.zip {url}{filename}'.format(
-            output=self.output().path, url=self.URL, filename=filename
-        ))
+        filenames = self._get_filename().splitlines()
+
+        for filename in filenames:
+            shell('wget -O {output}.zip {url}{filename}'.format(
+                output=self.output().path, url=self.URL, filename=filename
```
javitonino: 🤔 This is going to download two times into the same output file, is it no?

javitonino:
I think we need an smarter approach for the SP (São Paulo) case. We would need to download both files, unzip, and merge somehow. Or split this into two completely separate tasks (treat them as two separate states).

tasks/br/data.py
antoniocarlon: It's iterating over the output of `curl | awk | grep` to extract each line file name. For example the output of 
```
curl -s ftp://ftp.ibge.gov.br/Censos/Censo_Demografico_2010/Resultados_do_Universo/Agregados_por_Setores_Censitarios/ | awk '{print $9}' | grep SP_[A-Za-z_]*[0-9].*zip$
```
returns two lines 

```
SP_Capital_20150728.zip
SP_Exceto_a_Capital_20150527.zip
```
And the output of 
```
curl -s ftp://ftp.ibge.gov.br/Censos/Censo_Demografico_2010/Resultados_do_Universo/Agregados_por_Setores_Censitarios/ | awk '{print $9}' | grep ES_[A-Za-z_]*[0-9].*zip$
```

returns one line

```
ES_20161202.zip
```

javitonino: Yep, so the `wget` will run twice, writing to the same file (`output.zip`). So we will download `SP_Capital_20150728.zip`  first, and then overwrite it with `SP_Exceto_a_Capital_20150527.zip`.

it appears we would be losing the data from the capital file.

javitonino:
It turns out that SP has already being correctly handled, parsing it as two states. See https://github.com/CartoDB/bigmetadata/blob/master/tasks/br/geo.py#L88-L117. So I just fixed the regexp to avoid it picking up files that it shouldn't. 

Also I took a look at the updates file in the FTP and saw that Basico-MG has been fixed (albeit with a slightly different filename (dash instead of underscore)). I added it back to the ETL.

antoniocarlon:
Working, both in local and the ETL machine

PR #256, opened 2017-08-25
Added Canada targets to Makefile

antoniocarlon:
Added Canada NHS, census and geo targets to Makefile


javitonino:
Can we add a `ca-all` for all of Canada? It can just be a simple target depending on the others.

PR #257, opened 2017-08-25
Added Makefile targets for ES

antoniocarlon:
Added Makefile targets for ES

Also solves #259

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,7 +334,7 @@ es-ine-fyp:
 
 es-lacaixa:
 	docker-compose run --rm bigmetadata luigi \
-		--module tasks.es.lacaixa AnuarioWrapper \
+		--module tasks.es.lacaixa AnuarioAll \
 		--parallel-scheduling --workers=8
 
 ### eurostat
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -312,6 +312,31 @@ ca-geo:
...
+	docker-compose run --rm bigmetadata luigi \
+		--module tasks.es.ine PopulationHouseholdsHousingMeta \
+		--parallel-scheduling --workers=8
+
+es-ine-fyp:
+	docker-compose run --rm bigmetadata luigi \
+		--module tasks.es.ine FiveYearPopulationMeta \
+		--parallel-scheduling --workers=8
+
+es-lacaixa:
```
javitonino: There are some parameters missing for this one. We have loaded Anuarios for 2008, 2011 and 2013. I think we need to run this task multiple times, one per timespan. Hard to test though, because it is broken 😆

antoniocarlon: Yes, I'm planning to solve this in #259

Makefile
```diff
@@ -312,6 +312,31 @@ ca-geo:
...
+	docker-compose run --rm bigmetadata luigi \
+		--module tasks.es.lacaixa AnuarioWrapper \
```
javitonino: You forgot to update this Makefile target to use the new task ;)

PR #266, opened 2017-08-28
Added Makefile targets for EU

antoniocarlon:
Added Makefile targets for Eurostat


Makefile
```diff
@@ -338,9 +338,16 @@ es-lacaixa:
 		--parallel-scheduling --workers=8
 
 ### eurostat
-eurostat-data:
+eu-all: eu-geo eu-data
+
+eu-geo:
+	docker-compose run --rm bigmetadata luigi \
+	  --module tasks.eu.geo NUTSGeometries \
+	  --parallel-scheduling --workers=2
```
javitonino: We could update the number of workers for both tasks as well.

PR #277, opened 2017-08-29
Added US targets

antoniocarlon:
- Added US targets
 - Solved minor issues

### DIFF
--- a/tasks/us/census/acs.py
+++ b/tasks/us/census/acs.py
@@ -24,6 +24,14 @@ from tasks.tags import SectionTags, SubsectionTags, UnitTags, LicenseTags
 from time import time
 LOGGER = get_logger(__name__)
 
+GEOGRAPHIES = ['state', 'county', 'census_tract', 'block_group',
+                        'puma', 'zcta5', 'school_district_elementary',
+                        'congressional_district',
+                        'school_district_secondary',
+                        'school_district_unified', 'cbsa', 'place']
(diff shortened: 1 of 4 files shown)


tasks/us/census/acs.py
```diff
@@ -24,6 +24,14 @@
 from time import time
 LOGGER = get_logger(__name__)
 
+GEOGRAPHIES = ['state', 'county', 'census_tract', 'block_group',
+                        'puma', 'zcta5', 'school_district_elementary',
+                        'congressional_district',
+                        'school_district_secondary',
+                        'school_district_unified', 'cbsa', 'place']
```
javitonino: The alignment is a bit off here. Shouldn't everything be aligned so that continuation lines start at the same level as `'state'`?

tasks/us/census/acs.py
```diff
@@ -24,6 +24,14 @@
...
+YEARS = ['2015', '2010']
+SAMPLES = ['5yr', '1yr']
```
javitonino: This does not cover all cases, we have:
-5 years: 2010, 2015
-1 year: 2010, **2014**, 2015

tasks/us/census/acs.py
antoniocarlon: Blame the linter, but you are right, I'm changing it

javitonino: f*** linter!

Issue #260, opened 2017-08-25
Review FiveYearPopulation task for Spain


javitonino:
This task is adding a `gender` column to the metadata that is not used. We should remove that for sure.

We could also extract additional data from this file. Right now we only extract the total population (both genders), but we could add extra columns for males and females of each age group.

PR #283, opened 2017-08-30
Removed gender column and added separate columns for each gender

antoniocarlon:
- Removed gender column
- Added separate columns for each gender

#260

### DIFF
--- a/tasks/es/ine.py
+++ b/tasks/es/ine.py
@@ -16,6 +16,10 @@ from tasks.tags import SectionTags, SubsectionTags, UnitTags, BoundaryTags
 from time import time
 
 
+MALE = 'male'
+FEMALE = 'female'
+
+
 class DownloadGeometry(Task):
 
     URL = 'http://www.ine.es/censos2011_datos/cartografia_censo2011_nacional.zip'
@@ -1683,6 +1687,7 @@ class SeccionColumns(ColumnsTask):
         for _, col in columns.iteritems():
             col.tags.append(source)
             col.tags.append(license)
+
         return columns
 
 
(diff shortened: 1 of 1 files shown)


tasks/es/ine.py
```diff
@@ -1869,37 +1874,82 @@ def columns(self):
...
+        columns = OrderedDict()
+
+        genders = [None, MALE, FEMALE]
+
+        total_pop_male = OBSColumn(
+            id='total_pop_male',
             type='Numeric',
-            name='Population age 100 or more'.format(
-                start=start, end=end),
+            name='Total population (male)',
```
javitonino: Uhm, we have this column, with another name (`Males`) from INE data. How can we solve this duplication?

Ideas:
- Use two columns. Bad user experience.
- Use the same id for both columns. Since they have different boundaries, there is no conflict for now, could be in the future. If there is conflict, we would have to make sure the value in both datasets is the same.
- Mark the new columns with no tags. That should make them disappear from numerators, but appear on denominators (untested)

Rereading the code, it seems we are already doing approach 2,and it seems reasonable enough. Does that sound good?

antoniocarlon: Using no tags on the new columns makes them dissapear both from numerators and denominators (tested).
The current approach sounds reasonable enough for me.

javitonino: Ok, then let's go for it.

javitonino:
Uhm, now I confused. I though we agreed to use the columns from INE as denominators for the new data. But I see this back in CR with no changes (i.e: creates an additional column for the same concept). 😕

antoniocarlon:
Oops! yes, my bad

PR #285, opened 2017-08-30
253 uk fixes

### LINKED ISSUE #253: Fix and improve UK
- [x] Download links are broken (they have broken links in the [census webpage](https://www.ons.gov.uk/census/2011census/2011censusdata/bulkdata/bulkdatadownloads)). [Alternative link](http://webarchive.nationalarchives.gov.uk/20140115215021/http://data.statistics.gov.uk/Census/BulkLocalCharacteristicsoaandinfo310713.zip)
- [x] Add [Scotland](https://github.com/CartoDB/bigmetadata/issues/58) and Northern Ireland data. Note that there are some jobs for Scotland already in the code.
- [x] Add Total Male/Female population https://github.com/CartoDB/observatory-extension/issues/169
- [ ] Add geonames

javitonino:
#253 fixes, no improvements since this was getting quite big.

Highlights:
- ONS download was broken. There is an archived version, but the archive servers are very slow (20kb/s). I found an alternative source that has the same data, but with two levels of zips. I updated the task to handle that.
- CDRC download requires a login, and talos's cookie was outdated. I added an env variable to specify the cookie instead of hardcoding it.
- I also made the `ImportAllEnglandWalesLocal` task parallelizable. Before it returned subtasks one by one. Now it returns them all at once, so Luigi can schedule parallel execution of them
- Finally, a fix to some DB connection pooling problems. The problem is still there, I did not find a good solution to it after a few hours. So I just added a connectivity check before using a connection.
- Makefile targets

### DIFF
--- a/tasks/meta.py
+++ b/tasks/meta.py
@@ -783,6 +783,11 @@ class CurrentSession(object):
             print 'FORKED: {} not {}'.format(self._pid, os.getpid())
         if not self._session:
             self.begin()
+        try:
+            self._session.execute("SELECT 1")
+        except Exception as e:
+            self._session = None
+            self.begin()
         return self._session
 
     def commit(self):
--- a/tasks/meta.py
+++ b/tasks/meta.py
@@ -785,7 +785,7 @@ class CurrentSession(object):
             self.begin()
         try:
             self._session.execute("SELECT 1")
-        except Exception as e:
+        except:
             self._session = None
             self.begin()
         return self._session
--- a/tasks/uk/ons.py
+++ b/tasks/uk/ons.py
@@ -10,6 +10,7 @@ from tasks.tags import UnitTags, SectionTags, SubsectionTags, LicenseTags
 
 from collections import OrderedDict
 import os
+import shutil
 
 class SourceTags(TagsTask):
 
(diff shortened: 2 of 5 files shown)


tasks/meta.py
```diff
@@ -783,6 +783,11 @@ def get(self):
             print 'FORKED: {} not {}'.format(self._pid, os.getpid())
         if not self._session:
             self.begin()
+        try:
+            self._session.execute("SELECT 1")
+        except Exception as e:
```
antoniocarlon: We could remove `as e`

tasks/uk/ons.py
```diff
@@ -25,11 +26,21 @@ def tags(self):
...
+        shell('wget -O {output}.zip {url}'.format(output=self.output().path, url=self.URL))
+
+    def run(self):
+        super(DownloadEnglandWalesLocal, self).run()
+        work_dir = self.output().path
+        try:
+            for filename in os.listdir(work_dir):
+                if filename.endswith('.zip'):
+                    shell('unzip -d {path} {path}/{zipfile}'.format(path=work_dir, zipfile=filename))
+        except Exception as err:
```
antoniocarlon: We could remove `as err`

tasks/uk/ons.py
```diff
@@ -1119,9 +1130,15 @@ def run(self):
...
         for infile in infiles.strip().split('\n'):
-            table = os.path.split(infile)[-1].split('DATA.CSV')[0]
-            data = yield ImportEnglandWalesLocal(table=table)
+            tables.append(os.path.split(infile)[-1].split('DATA.CSV')[0])
+
+        datas = yield [ImportEnglandWalesLocal(table=table) for table in tables]
+
+        for data in datas:
             fhandle.write('{table}\n'.format(table=data.table))
         fhandle.close()
```
antoniocarlon: I would surround this with a `try-finally` or using `with`. What do you think?

Issue #282, opened 2017-08-30
make us-zillow fails


antoniocarlon:
Review the US Zillow data import

PR #288, opened 2017-08-31
Avoiding current month of current year as no data is available

antoniocarlon:
Solves #282

Avoiding current month of current year as no Zillow data is available


javitonino:
During acceptance, I found that we should also load the data for November and December of 2010 to keep backwards compatibility. Back to dev.

Issue #278, opened 2017-08-30
make us-spielman fails (Unfulfilled dependency at run time)


antoniocarlon:
`RuntimeError: Unfulfilled dependency at run time: DownloadSpielmanSingletonFile__99914b932b`

Related?: https://github.com/spotify/luigi/issues/1552

antoniocarlon:
It seems that the download URL and the form parameters to download the data have changed, rendering the `DownloadSpielmanSingletonFile` task useless.

We need to
 - Chek if this URL has the info we need https://www.openicpsr.org/openicpsr/project/100235/version/V5/view?path=/openicpsr/100235/fcr:versions/V5/Output-Data&type=folder
 - Rewrite the `DownloadSpielmanSingletonFile` task (if possible) or manually downloading the file and storing it (AWS?)

antoniocarlon:
If the data we need is [here](https://www.openicpsr.org/openicpsr/project/100235/version/V5/view?path=/openicpsr/100235/fcr:versions/V5/Output-Data&type=folder), the URL to download the data (change it [here](https://github.com/CartoDB/bigmetadata/blob/master/tasks/us/census/spielman_singleton_segments.py#L25-L34)) is https://www.openicpsr.org/openicpsr/project/100235/version/V5/download/project?dirPath=/openicpsr/100235/fcr:versions/V5/Output-Data

PR #295, opened 2017-09-01
Solving spielman import issues

antoniocarlon:
Solves #278

 - Changed download URL
 - Checked files and file renaming:
    - File `US_tract_clusters_new_20150720_401PM.shp` is incomplete
    - File `US_tract_clusters_new_20150720_427PM.shp` is complete and had to be renamed to `US_tract_clusters_new.shp` to match the rest of the Shapefile

### DIFF
--- a/tasks/us/census/spielman_singleton_segments.py
+++ b/tasks/us/census/spielman_singleton_segments.py
@@ -43,7 +43,7 @@ class DownloadSpielmanSingletonFile(Task):
         shell(cmd)
 
     def output(self):
-        return LocalTarget(path=os.path.join(classpath(self), self.filename()))
+        return LocalTarget(path=os.path.join('tmp', classpath(self), self.filename()))
 
 
 class ProcessSpielmanSingletonFile(Task):
@@ -94,10 +94,7 @@ class ProcessSpielmanSingletonFile(Task):
 
     def rename_files(self):
         rename = {
-            'ustractclustersnew-41530.bin' : 'US_tract_clusters_new.dbf',
-            'ustractclustersnew-41538.txt' : 'US_tract_clusters_new.prj',
-            'ustractclustersnew-41563.bin' : 'US_tract_clusters_new.shp',
-            'ustractclustersnew-41555.bin' : 'US_tract_clusters_new.shx'
+            'US_tract_clusters_new_20150720_427PM.shp': 'US_tract_clusters_new.shp',
         }
         for old_name, new_name in rename.iteritems():
             shell('mv {folder}/{old_name} {folder}/{new_name}'.format(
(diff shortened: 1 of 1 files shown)


tasks/us/census/spielman_singleton_segments.py
```diff
@@ -23,38 +23,33 @@ def filename(self):
...
+        return 'https://www.openicpsr.org/openicpsr/project/100235/version/V5/download/project?dirPath=/openicpsr/100235/fcr:versions/V5/Output-Data'
 
     def curl_request(self):
-        docid = 41528
-        email = 'slynn@cartodb.com'
-        return 'curl -L {url} --data "downloaderEmail={email}&agree=I Agree&id={docid}"'.format(
-            docid = docid,
-            email=email,
-            url= self.url());
+        return 'curl -L {url}'.format(url=self.url())
```
ethervoid: Why don't we use requests instead of calling curl and checking for system output?

tasks/us/census/spielman_singleton_segments.py
```diff
@@ -99,10 +94,7 @@ def convert_to_csv(self):
 
     def rename_files(self):
         rename = {
-            'ustractclustersnew-41530.bin' : 'US_tract_clusters_new.dbf',
-            'ustractclustersnew-41538.txt' : 'US_tract_clusters_new.prj',
-            'ustractclustersnew-41563.bin' : 'US_tract_clusters_new.shp',
-            'ustractclustersnew-41555.bin' : 'US_tract_clusters_new.shx'
+            'US_tract_clusters_new_20150720_427PM.shp': 'US_tract_clusters_new.shp',
```
ethervoid: Maybe we should rename the file to a custom name to avoid a name change (but not content) that could lead to this task to fail. I know is useful to know if the file has changed but we should rely on a content hash and that is added to the roadmap in the `Data repository`

tasks/us/census/spielman_singleton_segments.py
ethervoid: Even create a helper to download files using requests, if we don't have one, to be used as "our internal API"

antoniocarlon: We need to have a working ETL ASAP. I have created a new issue (#301) to solve this

ethervoid: Great :)

tasks/us/census/spielman_singleton_segments.py
antoniocarlon: This may work for most cases but it would fail now. The problem is that the downloaded zip contains  a malformed Shapefile: it contains a `prj`, a `shx` (which is correct), but two `shp` files, one partial and one total, and for both files the name is incorrect. In this case the only solution is to change the name of the `shp` file to match the rest of the Shapefile

ethervoid: Ok understood

ethervoid:
We should fix the tests

antoniocarlon:
Created task https://github.com/CartoDB/bigmetadata/issues/303 to fix tests

Issue #296, opened 2017-09-01
Extra column createdate in LODES


antoniocarlon:
Azure has extra column `us.census.lodes.createdate`

antoniocarlon:
This is a duplicate of #291

PR #298, opened 2017-09-01
Removed createdate column from LODES

### LINKED ISSUE #291: Extra column in us.census.lodes
We are adding a column named Creation Date to the lodes data which marks when the data point was created. That doesn't seem useful information, and should probably not be published.

antoniocarlon:
Fixes #291 (#296 is duplicaded)

Removed `createdate` column from LODES

### DIFF
--- a/tasks/us/census/lodes.py
+++ b/tasks/us/census/lodes.py
@@ -561,11 +561,6 @@ class WorkplaceAreaCharacteristicsColumns(ColumnsTask):
                 targets={total_jobs: 'denominator'},
                 tags=[tags['employment'], tags['commerce_economy']]
             )),
-            ('createdate', OBSColumn(
-                type='Date',
-                name='Date on which data was created, formatted as YYYYMMDD ',
-                weight=0
-            )),
         ])
         for colname, col in cols.iteritems():
             col.tags.append(source)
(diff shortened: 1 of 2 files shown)


tasks/us/census/lodes.py
```diff
@@ -635,7 +630,7 @@ def columns(self):
     def populate(self):
         for infile in self.input()['data']:
             # gunzip each CSV into the table
-            cmd = r"gunzip -c '{input}' | psql -c '\copy {tablename} FROM STDIN " \
+            cmd = r"gunzip -c '{input}' | cut -d',' -f-52 | psql -c '\copy {tablename} FROM STDIN " \
```
ethervoid: Again, add this to the helper #301 to avoid call system utils when we're able to do it with python. For example:
- gunzip could be done with gzip library
- `cut` could be done with split
- copy command could be achieved with psycopg2

tasks/us/census/lodes.py
```diff
@@ -561,11 +561,6 @@ def columns(self):
                 targets={total_jobs: 'denominator'},
                 tags=[tags['employment'], tags['commerce_economy']]
             )),
-            ('createdate', OBSColumn(
```
ethervoid: Probably we have to make a bump in the version field because we're changing columns

tasks/us/census/lodes.py
antoniocarlon: Updated https://github.com/CartoDB/bigmetadata/issues/301

tasks/us/census/lodes.py
antoniocarlon: Version bumped

ethervoid: Shouldn't be the version bump in the lodes columns?

antoniocarlon: I have bumped both the `ColumnsTask` and the `TableTask` version

ethervoid: I see the bump changes in the `spielman_singleton_segments.py` file :\

antoniocarlon: Oops... solved

Issue #303, opened 2017-09-04
Review DO tests


antoniocarlon:
Review DO tests after last changes

PR #305, opened 2017-09-05
Fix test error introduced by changes in Zillow

ethervoid:
We've introduced new data in zillow but for only two months. The
test cases uses the default parameters and in this case we're testing
against a case without data


antoniocarlon:
This solves #303

PR #312, opened 2017-09-12
Added logs for postgresql container

ethervoid:


docker-compose.yml
```diff
@@ -4,7 +4,6 @@
   "postgres": {
     "image": "carto/bigmetadata_postgres",
     "volumes": [
-      "./postgres/postgresql.conf:/etc/postgresql/9.5/main/postgresql.conf",
```
javitonino: If you remove it, how does it know to mount the postgres configuration file?

ethervoid: It [copies the file](https://github.com/CartoDB/bigmetadata/pull/312/files#diff-919c1cc4045b5a538c5600a717a3baabR26) when you create the image of the container. From my PoV having a shared configuration file that could change is not a good idea, containers should be efemeral if we need to change the config, we create a new image and that's it

javitonino: Oh, so COPY loads from outside the docker. I did not know that, I thought it was looking inside the image. 👍 then

ethervoid: Yep it does the copy operation at building time

PR #315, opened 2017-09-19
Fixes OOM during shoreline clip

### LINKED ISSUE #299: us.census.tiger fails
The target `make us-tiger` fails:

    File "/bigmetadata/tasks/us/census/tiger.py", line 302, in columns
      clipped[colname + '_clipped']._column: GEOM_NAME

Example: task `SplitSumLevel(year=2015, geography=county)`

javitonino:
This reengineers the ShorelineClip query, so it works row by row, instead of dumping the whole table and then regrouping it by id. The result is that psql only needs to keep a single row in RAM, instead of the whole table.

Closes #299

### DIFF
--- a/tasks/us/census/tiger.py
+++ b/tasks/us/census/tiger.py
@@ -666,14 +666,14 @@ class ShorelineClip(TableTask):
 
         stmt = ('''INSERT INTO {output}
                    SELECT
-                     geoid, aland,
+                     geoid,
                      ST_Union(ARRAY(
                        SELECT ST_MakePolygon(ST_ExteriorRing(
                          (ST_Dump(the_geom)).geom
                        ))
                      )),
-                   FROM {input}
-                   WHERE GeometryType(the_geom) = 'MULTIPOLYGON';'''.format(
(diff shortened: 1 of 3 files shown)


tasks/us/census/tiger.py
```diff
@@ -663,15 +663,18 @@ def timespan(self):
...
+                   SELECT
+                     geoid,
+                     ST_Union(ARRAY(
+                       SELECT ST_MakePolygon(ST_ExteriorRing(
+                         (ST_Dump(the_geom)).geom
+                       ))
+                       WHERE GeometryType(the_geom) = 'POLYGON'
+                     )),
+                     aland
+                   FROM {input}'''.format(
```
antoniocarlon: Both queries return the same results (checked with states and counties). Analyzing the queries with `EXPLAIN ANALYZE`, the query planner is using a `HashAggregate` for the old query and a `Seq Scan` for the new query, so the new query is working as expected (performing the memory consuming operations per row instead of storing the whole table in memory)

PR #321, opened 2017-09-20
Created ETL Task for the geometry simplification using mapshaper

### LINKED ISSUE #292: Geometry simplification proof of concept
From https://github.com/CartoDB/observatory-extension/issues/304

Summary of what we have learned:
- Our intersection algorithms are fine, the complexity is in the data
- We could simplify geometries to improve this:
  - User geometry would have to be simplified on-the-fly which is slow. Instead, we have added some limits to the pre-check to limit its effect.
  - DO boundaries could benefit for simplification
- We can simplify DO geometries to improve performance significantly. But we have some issues with topologies, since that can cause holes or overlaps
- Some benchmarking shows that the tradeoff could be worth it. We lose little precision and gain a lot in performance. DO is prepared to deal with the overlap and hole cases without throwing errors.
- We could reduce holes/overlaps by using some more advanced simplification techniques (without going to full topology generation). e.g: https://github.com/mbloch/mapshaper

Next steps:
- [x] Create an ETL task for geometry simplification
- [x] Run it for the data of one or more countries
- [x] Collect some usecases/maps using DO data. See how they behave with the simplified geometries

antoniocarlon:
Created ETL Task for the geometry simplification using mapshaper. Includes the following tasks:
 - **ExportShapefile**: Generates a Shapefile in a tmp directory from the input schema.table
 - **SimplifyShapefile**: Simplifies the temp shapefile using mapshaper (the retain percentage can be configured and is set to 50% by default)
 - **ImportSimplifiedShapefile**: Imports the simplified shapefile in the desired schema.table (the suffix of the table can be configured to avoid overwriting)
 - **SimplifyGeometries**: Wrapper for the tasks

Closes #292


tasks/simplification.py
```diff
@@ -0,0 +1,227 @@
...
+                                    schema=schema,
+                                    dont_simplify="" if self.dont_simplify is None
+                                                  else "AND id NOT SIMILAR TO '" + self.schema + ".("
+                                                  + "|".join([x for x in self.dont_simplify]) + ")%'"
+                                )).fetchall()
+
+    def find_simplification(self, tableid, tablename):
+        simplification = SimplifyGeometriesMapshaper(OBSERVATORY_SCHEMA, tablename)
+
+        if self.override_defaults is not None:
```
javitonino: You can make `override_defaults` be `[]` by default (instead of `None`) and remove this check.

tasks/simplification.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class Simplify(Task):
+    schema = Parameter()
+
+    dont_simplify = None
+    '''
+    Array of names of tables in the schema that we don't want to simplify
+    Example: dont_simplify = ['mytable1', 'mytable2']
+    '''
```
javitonino: Thinking on how to integrate this, I'm not sure if a table that simplifies everything that has not been skipped is a good idea. For example, in the US census, you would have to skip all the intermediate tables of the clipping part.

Also, most tasks typically work with a single table, and this one works with the entire schema. I think it can be tricky to create a `TableTask` that takes the input from this simplification task as an input.

Maybe I'm missing context. Can you provide an example on how to use this for one dataset?

tasks/simplification.py
```diff
@@ -0,0 +1,227 @@
...
+        shell(cmd)
+        self.executed = True
+
+    def output(self):
+        return PostgresTarget(self.schema, self.table)
+
+    def complete(self):
+        if not self.executed:
+            return False
+        return self.output().exists()
```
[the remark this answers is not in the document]
antoniocarlon: The destination table always exists (the task is overwriting it). If we just check it the table exists the task would never execute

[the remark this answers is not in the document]
javitonino: Uhm, I did not notice that. Wouldn't that be the case that each time that we run the parent task we would get a more simplified version each time? I mean, first time you simplify, next time, this will run again (because of the not executed part) and it would simplify the already simplified table, leading to even more simplified geos. Wouldn't it?

tasks/simplification.py
```diff
@@ -0,0 +1,227 @@
...
+    return '{schema}.{table}{suffix}.shp'.format(schema=schema, table=table, suffix=suffix)
+
+
+class ExportShapefile(Task):
+    schema = Parameter()
+    table = Parameter()
+    skipfailures = Parameter(default=SKIPFAILURES_NO)
+
+    def run(self):
+        self.output().makedirs()
```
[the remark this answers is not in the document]
antoniocarlon: The Shapefile is not only one file but a bunch of files. `temporary_path` generates a path like `tmp/simplification/es.ine.RawGeometry__99914b932b/RawGeometry__99914b932b.shp-luigi-tmp-4386797297` that `ogr2ogr` interprets as a directory rather than a file. If we use the generated temp directory for the output, `ogr2ogr` will use the table name as the file name, which is lowercase (for example `rawgeometry__xxx` instead of `RawGeometry__xxx`) and it breaks all the other tasks of the simplification unless we explicitly set the input parameters to be lowercase which is ugly

tasks/es/ine.py
```diff
@@ -122,6 +123,9 @@ def populate(self):
                             output=self.output().table,
                             input=self.input()['data'].table))
 
+    def post_tasks(self):
```
javitonino: What is `post_tasks`? I can't see where this is called.

tasks/simplification.py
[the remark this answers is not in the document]
javitonino: And what about just using a fixed name inside the directory? Something like `shapefile=os.path.join(temp_dir, 'simplify.shp')`. We have to modify the rest of tasks as well, to do the same.

Or take a look at `AtomicLocalFile` and see if it works. IIRC, the problem there was that the context manager (`with ...`) returned a file object rather than a path.

Just thinking because I'd like to impose a rule to use the atomic files/directories for new code. It can be more work in some circumstances, but it guarantees safety for parallel running workers.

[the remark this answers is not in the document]
javitonino: Also, if we are not going to be overwriting, maybe it makes sense to do a directory per-task, for consistency with the rest of the code. Right now, it seems we are sharing the directory, and putting files with different names inside.

tasks/simplification.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def output(self):
+        return LocalTarget(os.path.join(tmp_directory(self.schema, self.table_input),
+                                        shp_filename(self.table_output)))
+
+
+class SimplifyGeometriesMapshaper(Task):
+    schema = Parameter()
+    table_input = Parameter()
+    table_output = Parameter()
```
javitonino: `table_output` should probably be decided by this task. Tasks that call this one can check this task output to determine where the destination table is.

Maybe even just making this task inherit from `TempTableTask` which automatically names tables is enough (although that might not work since this task is in a different file than the rest of the import process).

tasks/es/ine.py
antoniocarlon: Removed (it was from a previous attemp using dynamic dependencies)

tasks/simplification.py
antoniocarlon: I would prefer keeping the `table_output` parameter because it allows us to execute the simplify task separately (for example from the command line) maintaining full control of the output table

javitonino: Uhm, fair point. What about adding a default then? That way we support both usecases. 

If this sounds good, give it a try. But only if it's easy to do, if there are problems with making one parameter default to depend on another value, let's leave it as it is.

tasks/simplification.py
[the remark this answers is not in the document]
antoniocarlon: Of course it makes sense addressing the atomic writes problem. I have modified it to use [the recommended `temporary_path()`](http://luigi.readthedocs.io/en/stable/luigi_patterns.html#atomicwrites).

As the Shapefile is formed by several files I have implemented the solution based on the directory containing the Shapefile. The main drawback is that the `output` for the task is a `LocalTarget` pointing to the directory containing the Shapefile and `temporary_path()` creates only the parent directories (not the Shapefile directory itself), so we need to manually create the directory for the Shapefile doing `self.output().fs.mkdir(temp_path)`. Besides that, the temporary directory (with Luigi managing the renaming and deleting tasks) and the contained Shapefile behave as expected

tasks/simplification.py
antoniocarlon: Done with a bit of magic (rewriting the `__init__` method) but the solution is effective and not _very_ ugly

javitonino:
There are some failing tests, not sure if relevant to the PR or just random failures. I'm relaunching the tests now, just in case.

tasks/simplification.py
```diff
@@ -0,0 +1,156 @@
...
+
+        columns = session.execute("SELECT column_name "
+                                  "FROM information_schema.columns "
+                                  "WHERE table_schema = '{schema}' "
+                                  "AND table_name   = '{table}'".format(
+                                    schema=self.schema, table=self.table_input.lower())).fetchall()
+
+        factor = postgis_simplification_factor(self.schema, self.table_input, self.geomfield, self.retainfactor)
+
+        simplified_geomfield = 'ST_MakeValid(ST_SimplifyVW({geomfield}, {factor})) {geomfield}'.format(
```
ethervoid: One question, we should not make here a ST_CollectionExtract(...,3) in order to get only polygon/multipolygon instead of geometrycollection which we don't support?

antoniocarlon: The idea of these tasks is simplifying the data of an existing table (an existing `TempTableTask`) so I wouldn't expect unsupported geometry types, but as we are using the `-nlt MULTIPOLYGON` option for the `ogr2ogr` import, we could add `ST_CollectionExtract(...,3)` as you suggest for symmetry.

ethervoid: My question was more in the line that `ST_Makevalid` could create `GeometryCollection` from the validated geometries as it says:

```In case of full or partial dimensional collapses, the output geometry may be a collection of lower-to-equal dimension geometries or a geometry of lower dimension.```

I've suffered from that when I have to make geometries valid. I don't know if that affect us here but just in case I prefer to ask hehehe :)

antoniocarlon: Oh... yes! You're right! I'm updating it :)

PR #322, opened 2017-09-20
320 geoname nulls

### LINKED ISSUE #320: ShorelineClip geos have no geonames
And tests fail

javitonino:
Fixes #320

### DIFF
--- a/tasks/us/census/tiger.py
+++ b/tasks/us/census/tiger.py
@@ -504,7 +504,7 @@ class SplitSumLevel(TempTableTask):
             output=self.output().table))
 
     def original_task(self):
-        self.requires()
+        return self.requires()
 
 
 class JoinTigerWaterGeoms(TempTableTask):
@@ -607,7 +607,7 @@ class PreunionTigerWaterGeoms(TempTableTask):
             output=self.output().table))
 
     def original_task(self):
-        self.requires()['split'].original_task()
+        return self.requires()['split'].original_task()
 
 
 class UnionTigerWaterGeoms(TempTableTask):
@@ -632,7 +632,7 @@ class UnionTigerWaterGeoms(TempTableTask):
                             output=self.output().table,
                             input=self.input().table))
     def original_task(self):
-        self.requires().original_task()
+        return self.requires().original_task()
 
 
 class ShorelineClip(TableTask):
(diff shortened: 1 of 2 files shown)


tasks/us/census/tiger.py
```diff
@@ -503,6 +503,9 @@ def run(self):
         session.execute('CREATE INDEX ON {output} USING GIST (the_geom)'.format(
             output=self.output().table))
 
+    def original_task(self):
+        return self.requires()
```
ethervoid: Why is this a better option than passing the name through the queries which are related with the geoid? I think that create tables with the needed fields is the regular way to go. 

I know I'm a pain in the ass with this part but we're passing for example aland and awater through all the queries and in order countries we're doing the same so I want to be sure we're doing this "new thing" for a good reason

tasks/us/census/tiger.py
```diff
@@ -651,34 +660,48 @@ def requires(self):
...
+                   input=self.input()['data'].table,
+                   additional_columns=additional_columns,
+                   original_table=self.original_task().output().table)
         session.execute(stmt)
 
+    def original_task(self):
+        return self.requires()['data'].original_task()
+
+    def name(self):
+        return self.original_task().name
```
ethervoid: I know this is one of the reasons to use the list of dependencies to get the "original task" but we could have a data structure or POPO which links name column and the geography level

tasks/us/census/tiger.py
javitonino: We can propagate through the tables. This just felt easier to do during development, as you only need to alter one query depending on if the dataset has name or not, instead of all intermediate ones. Also, it's a bit faster, since the intermediate tables are smaller (and fit better in cache), but the effect is negligible (a few seconds).

I'm not opposed to changing this to propagate name through the needed parts of the query. However, we still need to propagate whether the task needs a name or not, either via requires (similar to this) or a global configuration dict.

tasks/us/census/tiger.py
javitonino: This is actually needed to get the original table name as well. If we remove that and go for propagating `name` through intermediate tables, I agree that it's easier to change this to a global configuration dict rather than going up the dependency chain.

ethervoid: My concern is to couple our code with the internal functions of Luigi. Deal with dependencies and core things of Luigi, even if they're available, is kind of bad smell. We should isolate and minimize this kind of things.

PR #329, opened 2017-09-26
Remove universe column relationship

### LINKED ISSUE #326: 'universe' target should not be an option for normalization
There are two types of "targets" in bigmetadata; universe and denominator. The denominator targets indicate how the variable can be normalized. The universe targets captures the variable used to calculate the median or per capita measurement. We should remove universe targets from the normalization option. We might also just remove the universe target entirely, because is there any known use for it? I'm not sure.

image

javitonino:
Closes #326

Remove all reference to universe

### DIFF
--- a/tasks/fr/fr_income.py
+++ b/tasks/fr/fr_income.py
@@ -146,17 +146,8 @@ class IrisIncomeColumns(ColumnsTask):
     def get_targets(self, cols, denominators):
         targets_dict = {}
         if denominators:
-            targets_dict = self.get_relation_targets(cols, denominators, DENOMINATOR)
-
-        return targets_dict
-
-    def get_relation_targets(self, cols, relations, relation):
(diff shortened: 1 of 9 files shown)


tasks/fr/fr_income.py
```diff
@@ -144,12 +143,10 @@ def columns(self):
 
         return cols
 
-    def get_targets(self, cols, denominators, universes):
+    def get_targets(self, cols, denominators):
         targets_dict = {}
         if denominators:
             targets_dict = self.get_relation_targets(cols, denominators, DENOMINATOR)
```
antoniocarlon: `get_relation_targets` was useful when we had both denominators and universes. Since we only have denominators now, we may consider removing the method

javitonino:
Merge master due to luigi upgrade.

PR #335, opened 2017-09-29
Fixing value name. Adding updates

### LINKED ISSUE #310: wrong value naming for total popultation raw data
I think it is confusing naming `per_sq_km` the value:

image

This should be fixed.

antoniocarlon:
Fixing value name (`per_sq_km`)
Adding `UPDATE` queries

Closes #310

### DIFF
--- a/catalog/templates/column.html
+++ b/catalog/templates/column.html
@@ -54,7 +54,7 @@ Code example
 
           WITH ids AS (
           SELECT ARRAY_AGG(<id_col>) ids
-            FROM <table_to_update>
+            FROM <table_with_ids>
           ),
           meta AS (
             SELECT OBS_GetMeta(ST_MakeEnvelope(-179, 89, 179, -89, 4326),
@@ -95,7 +95,7 @@ Code example
           FROM OBS_GetData(
                 (SELECT ARRAY_AGG((the_geom, id)::geomval) FROM data),
                 (SELECT meta FROM meta)) obs_data
-          WHERE <table_to_update>.<id_col> = obs_data.id
+          WHERE <table_to_update>.cartodb_id = obs_data.id
 
       Obtain an estimate of "{{ col.name }}" per {{ denom_name }} for a polygon
       using a combination of weighting by area and {{ denom_name }}.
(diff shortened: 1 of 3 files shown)


catalog/templates/column.html
```diff
@@ -192,16 +277,36 @@
                "normalization": "predenominated"}]'
             , 1, 1) meta
           FROM data)
-          SELECT id, (data->0->>'value')::{{ col.type }} as per_sq_km
+          SELECT id, (data->0->>'value')::{{ col.type }} as raw_count
```
javitonino: Uhm, I'm not sure if it's available in the RSTs, but the optimal solution would be to use the same column name we autogenerate for Builder, i.e: the `suggested_name` column of the GetMeta result. 

If not, we can probably add it to the query to generate rsts, or replicate this logic in python:
https://github.com/CartoDB/observatory-extension/blob/develop/src/pg/sql/41_observatory_augmentation.sql#L249-L257

antoniocarlon:
Putting the `UPDATE` before the `SELECT` was my first idea because nobody needs to only see the data but do something with it. I leaved the `SELECT` first to be more conservative and trying not to surprise the user but it's clear that we agree that putting the `UPDATE`s first is more natural so I'm changing it.

javitonino:
For reference, this is how this looks:

screenshot_20171002_102515

Some questions/suggestions:
- For updating a table, the `WITH data` part should probably take data from a table, something like:
```
WITH data AS (
  SELECT cartodb_id as id, the_geom FROM <table_to_update>
)
```
- We can probably replace `<id_col>` with `cartodb_id`.
- Maybe we can replace `<field_to_update>` with the suggested name. This feels like it may be too much.
- Is it possible to use the suggested name in all tabs? In Builder this works and shows names like `male_pop_per_sq_km_2015`

antoniocarlon:
Answeriing your suggestions:

1. Totally. The query with a fixed point doesn't make sense at all
2. 99% of the time `<id_col>` will be `cartodb_id`. I'm changing it
3. I prefer keeping the `<field_to_update>` placeholder. The suggested name is great for the `SELECT` as it gives you context about the result, but for some users it may be hard to infere that they need a column `pop_18_25` (for example) in their table
4. I think it's possible and it would be nice. I'll take a look

javitonino:
Very nice!

screenshot_20171002_134212

One minor nitpick before letting this go. We should capitalize `AS` in `SELECT cartodb_id as id, the_geom FROM <table_to_update>`, for consistency. After that, this is a 🚀

Issue #325, opened 2017-09-21
Some variables are always null at small geometries


javitonino:
At least the following variables have all null values for `block_group` geometries. Probably many more:
us.census.acs.C24050014
us.census.acs.B23006001
us.census.acs.B11009003
us.census.acs.B05001006
us.census.acs.C24050002
us.census.acs.B23006001
us.census.acs.B08201006
us.census.acs.B05001006

Also, the BLS ones that has suffice `_1028` or `_92`. These seem to have been manually deleted from previous dumps.

We didn't notice before because autotest check a single geometry level. Suggestions:
- Add a step at TableTask post-process that checks if a column only has NULLs (or a high %), and throw an error
- Review ACS/BLS data, and remove empty columns (or replace by zero, if that makes sense).
- Improve autotests to check against multiple/all geometry levels.

PR #337, opened 2017-10-04
Removing null columns

antoniocarlon:
#325 

- Added a TableTask check to avoid null columns
- Removed null columns on us.acs
- Removed null columns on us.bls


javitonino:
Some queries I found useful:

1. Analyze all the things:
```
DO
$do$
DECLARE
    tname text;
    tableid text;
BEGIN
FOR tableid, tname IN
    SELECT id, 'observatory' || '.' || tablename
    FROM   observatory.obs_table
    WHERE id LIKE '%acs%'
    ORDER BY id
LOOP
    RAISE NOTICE '%', tableid;
    EXECUTE 'ANALYZE ' || tname;
END LOOP;
END
$do$;
```

2. Get columns with 100% NULLs (after having completed an ANALYZE)
`SELECT id, array_agg(attname ORDER BY attname) FROM pg_stats JOIN observatory.obs_table USING (tablename) WHERE schemaname = 'observatory' AND null_frac = 1 GROUP BY tablename, id ORDER BY id;`

tasks/us/naics.py
```diff
@@ -2482,13 +2483,20 @@ def is_sector(code):
...
     '''
     Returns True if this is a NAICS supersector, False otherwise
     '''
     return len(code) == 4 and code[0:2] == '10'
 
 
+def is_public_administration(code):
+    if code in ('1028', '92'):
+        return True
+    return False
```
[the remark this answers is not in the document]
antoniocarlon: :man_facepalming:

tasks/util.py
```diff
@@ -1634,6 +1636,24 @@ def create_geom_summaries(self, output):
         generate_tile_summary(current_session(),
                               output._id, colid, output.table, colname)
 
+    def check_null_columns(self):
+        session = current_session()
+        result = session.execute('SELECT {values} FROM {table}'.format(
+                            table=self.output().table,
+                            values=', '.join(['sum(case when {column} is null then 0 else 1 end) {column}'.format(
+                                column=x) for x, _ in self.columns().iteritems()])))
```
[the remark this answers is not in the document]
javitonino: Since you went for analysing the table, you can do this very quickly with `pg_stats`

`SELECT attname FROM pg_stats WHERE schemaname = 'observatory' AND tablename = '{table}' AND null_frac = 1;`

tasks/util.py
```diff
@@ -1587,6 +1587,13 @@ def run(self):
         after = time.time()
         LOGGER.info('time: %s', after - before)
 
+        LOGGER.info('analyzing the table %s', self.output().table)
+        session = current_session()
+        session.execute('ANALYZE {table}'.format(table=self.output().table))
```
ethervoid: Depending in the size of the table this could take some time to finish

ethervoid: Maybe we should move the analyze inside the `check_null_columns` because is a dependency for that module to avoid removing by mistake or something

javitonino: I tested it, and it was a little over a minute for the slowest datasets (ACS block groups), faster than the old manual stats code. I suggested this, as it might be useful to have this information and it is not too slow.

antoniocarlon: I agree that, even if we remove the nulls check, having good statistics for the tables can be useful. Thus, although `check_null_columns` now depends on the `ANALYZE` I would leave the `ANALYZE` outside the function

javitonino:
Things (outside of code) to review:
- Tests are failing because of this check. Probably because some wrappers just import a fake table with no data?
- Some missing columns in other datasets:
  - Eurostat, a lot of `_flag` columns. We should remove the `_flag` fields, which only indicate if there is data for the associated field. They have also weight=0 so they do not show in output.
  - MX, some fields for low level geographies, similar to ACS.

tasks/us/census/acs.py
```diff
@@ -3250,12 +3155,124 @@ def columns(self):
...
+            ("pop_5_years_over", pop_5_years_over),
+            ("speak_only_english_at_home", speak_only_english_at_home),
+            ("speak_spanish_at_home", speak_spanish_at_home),
+            ('speak_spanish_at_home_low_english', speak_spanish_at_home_low_english),
+            ("pop_15_and_over", pop_15_and_over),
+            ("pop_never_married", pop_never_married),
+            ("pop_now_married", pop_now_married),
+            ("pop_separated", pop_separated),
+            ("pop_widowed", pop_widowed),
+            ("pop_divorced", pop_divorced),
```
javitonino: This is ok for now, but in the future we can probably migrate all of this to the `columns.json` file. It is quite easy to autogenerate (basically, iterate over this object and generate JSON about it). The only trick is to handle the references between columns (but we can copy the approach that I use in the UK thingy).

Anyway, we can leave it as it is for now, as this PR is getting big, and revisit in the future.

tasks/columns.py
```diff
@@ -0,0 +1,60 @@
+import json
+from collections import OrderedDict
+
+
+CONDITIONS = "conditions"
+EXCEPTIONS = "exceptions"
+
+
+class ColumnsDeclarations:
```
javitonino: Probably a bit ambitious to put this in the main directory for now. I would move it inside `us.census` for now, as it is only used there at the moment. Once we have the requirements for this metamagic a bit more clear, we can work in a generic approach. Probably after merging this and UK.

tasks/columns.py
```diff
@@ -0,0 +1,60 @@
...
+        for requirement in requirements:
+            check = True
+            for id, value in parameters.iteritems():
+                try:
+                    if requirement[id] != value:
+                        check = False
+                except:
+                    pass
+            if check:
+                return True
```
javitonino: Some ideas to simplify:

Just use `if id in requirements and requirements[id] != value` instead of the try/catch.

I think you can just do `return False` in the requirement, and `return True` after the loop, and remove the `check` variable entirely.

Bonus trick, have you heard about `for else`? https://docs.python.org/2/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops

Also, avoid using `id` since it is a Python built-in.

tasks/columns.py
```diff
@@ -0,0 +1,60 @@
...
+
+        return False
+
+    def filter_columns(self, columns, parameters):
+        filtered = OrderedDict()
+        for colid, value in columns.iteritems():
+            if self._is_column_filtered(colid, parameters):
+                filtered[colid] = value
+        return filtered
```
javitonino: I like list comprehensions for this:
`return OrderedDict([[k, v] for k, v in columns.items() if self._is_column_filtered(k, parameters)])`

tasks/columns.py
```diff
@@ -0,0 +1,60 @@
...
+
+        conditions = self._get_requirements(colid, CONDITIONS)
+        if conditions is not None:
+            if not self._check_requirements(params, conditions):
+                return False
+
+        exceptions = self._get_requirements(colid, EXCEPTIONS)
+        if exceptions is not None:
+            if self._check_requirements(params, exceptions):
+                return False
```
javitonino: Maybe you can skip `is not None`:
```
if None -> false
if {} -> false
if {'a':1} -> true
```

And put everything in the same line:
```
if exceptions and self._check_requirements(params, exceptions):
  return False
```

tasks/columns.py
```diff
@@ -0,0 +1,60 @@
...
+            self._columns = json.load(file)
+
+    def _find_column(self, colid):
+        return self._columns[colid]
+
+    def _get_requirements(self, colid, requirement):
+        try:
+            return self._find_column(colid)[requirement]
+        except:
+            return None
```
javitonino: Just do `return self._find_column(colid).get(requirement)` which is the same without the `try` block.

tasks/us/census/acs.py
antoniocarlon: That is the idea, we should merge the `columns.py` from this PR with [your column generation through definition](https://github.com/CartoDB/bigmetadata/blob/ab24f3b52b67fc601991d5a7d3f980dc15fa368c/tasks/uk/census/metadata.py#L68-L89). This would be the first step towards declarative definition of columns.

tasks/columns.py
antoniocarlon: Of course it's ambitious! ;) Ok, I'm moving this inside `us.census` until we discuss the declarative approach

tasks/us/census/acs.py
javitonino: My idea with UK was to serve as a prototype for now. It has some ugly things that are not ideal for reuse (for example, the way it handles summing of fields, which is a feature which probably is not useful for other countries).

I'm starting a document to talk about this, and see what parts we can do with a generic approach (probably, metadata generation) and which need to be done with per-country extensions.

tasks/columns.py
antoniocarlon: :heart: the `for else`

javitonino:
Did you check [my previous comment](https://github.com/CartoDB/bigmetadata/pull/337#issuecomment-335511353)? Tests are still failing.

antoniocarlon:
Checking for NULL columns during tests doesn't make sense as they use empty or fake data.
I have added two new issues (#348 and #348) to address the NULL columns found at Eurostat and MX.

javitonino:
👍 , moving to acceptance.

javitonino:
Uhm, we have a small problem here. The task is failing because conflicts between transactions of different ColumnsTask executions.

Before, there was a single ColumnsTask instance, so it only ran once, and so, there wasn't any problem. Now, the task has different parameters, so it will be run once per parameter combination. If two runs happen at the same time, there is a conflict with DB transactions, and it fails.

Ideas:
- Keep a single ColumnsTask that creates all columns, but has a method for returning the relevant ones for the parameter combination. We can use `significant=False` in the parameters (to mark them as not affecting the number of executions needed).
- User [resources](http://luigi.readthedocs.io/en/stable/configuration.html#resources) to limit the number of ColumnsTask that can run in parallel. This is flexible for more tasks. We could create a resource called `obs_column` with a single unit available, and mark the ColumnsTask as requiring one unit of this resource. This way, only one ColumnsTask will be executed at a time.

I like the second one, except for the fact that it relies in luigi server configuration, which feels a bit weird. But it seems to be Luigi's mechanism to handle this kind of issues. The first alternative is just in our code, but I think it could get complicated quickly.

antoniocarlon:
I totally agree, as this is a problem that arised from this PR we should address it (at least it deserves to take a look at the possible solutions)

javitonino:
Another issue. This makes metadata generation a lot slower. In particular, the denom5 query, which does a `ST_Union` for each combination of geometries available for a numerator. This has been greatly increased by this PR, since there is a lot more of combinations (some numerators are only available on certain geometries), so this explains the slowness.

```
2017-10-16 14:03:48,082 time taken for denom:5: 1328.51
INFO: time taken for denom:5: 1328.51
```

Anyway, this takes over 20minutes, so we should optimize it somehow.
This is the query: https://github.com/CartoDB/bigmetadata/blob/master/tasks/carto.py#L648-L668
BTW, the index doesn't appear to be used (it goes for a full scan).

antoniocarlon:
The `ColumnsTask` parallel execution issue was resolved using Luigi's `resources` on all the `ColumnsTask`s of ACS, relying on unespecified resources (not defined in the Luigi config file) as [the documentation](http://luigi.readthedocs.io/en/stable/configuration.html#resources) states that _Unspecified resources are assumed to have limit 1_.

We will need to keep an eye on this kind of error to see  if it's worth to extend this solution to all the targets moving the `resources` from the ACS's `Columns` and `QuantileColumns` classes to the general `ColumnsTask`

PR #343, opened 2017-10-10
253 improved uk

### LINKED ISSUE #253: Fix and improve UK

javitonino:
Fixes #253

Totally overhauls UK.

**Notable changes**
- Split ETL in multiple files
  - `ons.py` loads data from ONS and `scotland.py` from Scotland census
  - `metadata.py` parses `columns.json` and generates metadata. I modified `ColumnsTask` to be able to override the ID so we can keep it.
  - `wrapper.py` creates the final table, using the metadata to determine which sub-census to load data from.
- `columns.json` for declarative column definition
- New datasources for most data. See the linked issue for more information. We try to pull from UK-level table when possible. When it is not, we pull from specific per-census tables and join them.
- Try to avoid using `shell`

**Still pending** (but I want a first round of CR first)
- Remove some `shell` comands.
- Improve Scotland insertion. It uses [`session.execute`](http://docs.sqlalchemy.org/en/latest/orm/session_api.html#sqlalchemy.orm.session.Session.execute), which calls [`psycopg2.executemany`](http://initd.org/psycopg/docs/cursor.html#cursor.executemany) which is slow (it does not do batching for some reason). Instead, replace it with a native psycopg2 `COPY FROM` as shown in `ImportUK`.
- Abstract class for subtasks (abstract `id_to_column` method)

**Comments**
The `Census` task is a massive SQL generation thingy. It could be divided into smaller parts using UPSERT's and run it for each column/table (old strategy). That way the code is a bit simpler, but it still doesn't hide the complexities of joining from different columns/tables into the same one. Another approach would be to do specific tasks to generate more intermediate tables for this.

Anyway, just to say, I know it is complicated, but it's going to be that way one way or another (although it could be divided in smaller chunks). I wrote some docs in the function to help follow it, let's see if it is enough. If not, no worries, we can surely make this simpler.

**Before deploy**
Add Scotland and NISRA to https://carto.com/docs/carto-engine/data/license/


tasks/uk/census/metadata.py
```diff
@@ -0,0 +1,82 @@
...
+    def version(self):
+        return 5
+
+    def prefix(self):
+        return 'uk.ons'
+
+    def columns(self):
+        input_ = self.input()
+        sources = input_['source'].values()
+        license = input_['license']['uk_ogl']
```
ethervoid: We should start changing license for other name because in Python `license` is a reserved word

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+
+class DownloadUK(Task):
+    API_URL = 'https://www.nomisweb.co.uk/api/v01/dataset/def.sdmx.json?search={}*'
+    DOWNLOAD_URL = 'https://www.nomisweb.co.uk/api/v01/dataset/{id}.bulk.csv?time=2011&measures=20100&geography={geo}'
+
+    table = Parameter()
+
+    def run(self):
+        # Query API, extract table ID from name
+        meta = json.load(urllib.urlopen(self.API_URL.format(self.table)))
```
ethervoid: :_) at last using a library to make request. Why don't use requests?

javitonino: Why? In this case, this is very simple with the standard lib, so no need to pull an extra dependency. Of course, if we need to do something more complex (headers, cookies, ...) we could add it. But for this, I think `urllib` is fine.

tasks/uk/census/metadata.py
javitonino: Uhm, yeah, we can make this one be the first one :)

tasks/uk/census/ons.py
ethervoid: If we're going to add that dependency, which I'm pretty sure we're going to do because we have complex requests, why don't start using it here?

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+        api_id = (meta['structure']['keyfamilies']['keyfamily'][0]['id']).lower()
+
+        # Download for SA (EW,S) and OA (NI) in a single file
+        with self.output().temporary_path() as tmp:
+            os.mkdir(tmp)
+            with open(os.path.join(tmp, '{}.csv'.format(self.table)), 'w') as outcsv:
+                skip_header = False
+                for geo in ['TYPE258', 'TYPE299']:
```
ethervoid: Why don't extract this to a constant?

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+
+class ImportUK(TempTableTask):
+    table = Parameter()
+
+    def requires(self):
+        return DownloadUK(self.table)
+
+    @staticmethod
+    def id_to_column(colid):
+        return re.sub(r'[:/, \-\.\(\)]', '_', '_'.join(colid.split(';')[0].split(':')[-2:]))
```
ethervoid: It's better to compile the regexp when you're going to use more than once

ethervoid: Never mind https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile#452143 since 2.5 it compiles and caches it automatically

javitonino: Oh, I didn't know that it compiles automatically. Anyway, it's probably better to extract, for clarity.

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+            # We are faking the IDs, because Scotland bulk downloads uses the column name instead of the ID
+            datacols = [self.id_to_column(x) for x in header[3:]]
+
+            session = current_session()
+            with session.connection().connection.cursor() as cursor:
+                cursor.execute('CREATE TABLE {output} (date TEXT, geography TEXT, geographycode TEXT PRIMARY KEY, {cols})'.format(
+                    output=self.output().table,
+                    cols=', '.join(['{} NUMERIC'.format(c) for c in datacols])
+                ))
+                csvfile.seek(0)
```
ethervoid: I've seen that is needed to copy from table to file but from file to table?

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+
+                cursor.copy_expert(
```
ethervoid: Nice improvement despite of the awful way we need to get the cursor xD

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+                        table=self.output().table),
+                    csvfile)
+
+
+class DownloadEnglandWalesLocal(DownloadUnzipTask):
+
+    URL = 'https://www.nomisweb.co.uk/output/census/2011/release_4-1_bulk_all_tables.zip'
+
+    def download(self):
+        shell('wget -O {output}.zip {url}'.format(output=self.output().path, url=self.URL))
```
ethervoid: God has killed a puppy :(

javitonino: It's in the "still pending". I will fix it for sure.

tasks/uk/census/ons.py
javitonino: It's an ugly thing because I use this same descriptor with a `csv.reader` to read the header. I think the better option is to open a separate descriptor for the `copy` command.

tasks/uk/census/ons.py
javitonino: There is https://github.com/jmcarp/sqlalchemy-postgres-copy but it only works with SQLAlchemy models. :(

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+        return DownloadEnglandWalesLocal()
+
+    @staticmethod
+    def id_to_column(colid):
+        return colid
+
+    def run(self):
+        session = current_session()
+        infile = os.path.join(self.input().path, self.table + 'DATA.CSV')
+        headers = shell('head -n 1 {csv}'.format(csv=infile))
```
ethervoid: We've to change this for csv reader and get the first next

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+        cols = ['{} NUMERIC'.format(h) for h in headers.split(',')[1:]]
+        session.execute('CREATE TABLE {output} (GeographyCode TEXT, {cols})'.format(
+            output=self.output().table,
+            cols=', '.join(cols)
+        ))
+        session.commit()
+        shell("cat '{infile}' | psql -c 'COPY {output} FROM STDIN WITH CSV HEADER'".format(
```
ethervoid: So here we have to do the same than before with copy

tasks/uk/census/scotland.py
```diff
@@ -0,0 +1,56 @@
...
+            cols = [ImportScotland.id_to_column(c) for c in header[1:]]
+
+            session = current_session()
+            session.execute('CREATE TABLE {output} (GeographyCode TEXT PRIMARY KEY, {cols})'.format(
+                output=self.output().table,
+                cols=', '.join(['{} NUMERIC'.format(c) for c in cols])
+            ))
+
+            rows = []
+            for r in reader:
```
ethervoid: Why don't we remove the `S92` lines and then use COPY instead of INSERT? How many lines are we talking about?

tasks/uk/census/wrapper.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def columns(self):
+        cols = OrderedDict()
+        input_ = self.input()
+        cols['GeographyCode'] = input_['geom_columns']['oa_sa']
+        cols.update(input_['data_columns'])
+        return cols
+
+    def source_tables(self):
+        return set(itertools.chain(*[[d['table'] for d in col['data']] for col in COLUMNS_DEFINITION.values()]))
```
ethervoid: I'd simplify this expression doing more lines. Is very hard to read with multiple nested fors

tasks/uk/census/scotland.py
javitonino: Yeah, it's in the plans. I initially thought that `execute` will batch, but due to how psycopg is implemented, it doesn't 😞.

We don't even need to remove S92, since it is filtered later, so this will eventually become a simple COPY.

tasks/uk/census/wrapper.py
javitonino: Nah, this is perfectly readable xDD. Yeah, two nested list comprehensions is too much.

tasks/uk/census/scotland.py
```diff
@@ -0,0 +1,56 @@
...
+                if r[0].startswith('S92'):
+                    continue  # Skip contry-level summary
+
+                binds = {cols[i]: (0 if v == '-' else v.replace(',', '')) for i, v in enumerate(r[1:])}
```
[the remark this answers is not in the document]
javitonino: I like this and consider it pretty readable. This is how a lot of Python code looks, and I consider this to be decent (single loop, and only a bit of uglyness with the ternary conditional). Of course, in cases like the one below, that is definitely too much.

I can review it and see if I can make it cleaner, which I think I can.

tasks/uk/census/wrapper.py
```diff
@@ -0,0 +1,132 @@
...
+        return self.REGION_MAPPING[region](table=table)
+
+    def populate(self):
+        '''
+        Joins data from all UK sub-census dependencies.
+
+        For each column, there are multiple possiblities:
+          - The column has a single datasource and field: ``"data" : [{"table": "t", "fields": ["f"]}]``.
+            The table will be added to the FROM, and the column to the SELECT
+            ``SELECT t.f AS id FROM ... FULL JOIN t USING (geographycode)``
```
ethervoid: For a better understanding I'd change `id` for `out_data` or something like that

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,125 @@
...
+        meta = json.load(urllib.urlopen(self.API_URL.format(self.table)))
+        api_id = (meta['structure']['keyfamilies']['keyfamily'][0]['id']).lower()
+
+        # Download for SA (EW,S) and OA (NI) in a single file
+        with self.output().temporary_path() as tmp:
+            os.mkdir(tmp)
+            with open(os.path.join(tmp, '{}.csv'.format(self.table)), 'w') as outcsv:
+                skip_header = False
+                for geo in ['TYPE258', 'TYPE299']:
+                    remote_file = urllib.urlopen(self.DOWNLOAD_URL.format(id=api_id, geo=geo))
```
antoniocarlon: We could use `with` here to ensure that `remote_file` is closed

tasks/uk/census/scotland.py
```diff
@@ -0,0 +1,56 @@
...
+        shell('wget -O {output}.zip {url}'.format(output=self.output().path, url=self.URL))
+
+
+class ImportScotland(TempTableTask):
+    table = Parameter()
+
+    def requires(self):
+        return DownloadScotlandLocal()
+
+    @staticmethod
```
antoniocarlon: Why using a static method instead of a top level function?

tasks/uk/census/scotland.py
```diff
@@ -0,0 +1,56 @@
...
+
+            session = current_session()
+            session.execute('CREATE TABLE {output} (GeographyCode TEXT PRIMARY KEY, {cols})'.format(
+                output=self.output().table,
+                cols=', '.join(['{} NUMERIC'.format(c) for c in cols])
+            ))
+
+            rows = []
+            for r in reader:
+                if r[0].startswith('S92'):
```
antoniocarlon: We could extract this to a constant and maybe hide the implementation details creating a `is_importable` function

tasks/uk/census/metadata.py
```diff
@@ -0,0 +1,82 @@
...
+            'license': LicenseTags(),
+        }
+
+    def version(self):
+        return 5
+
+    def prefix(self):
+        return 'uk.ons'
+
+    def columns(self):
```
antoniocarlon: I :heart: this method. This should be everywhere

tasks/uk/census/wrapper.py
```diff
@@ -0,0 +1,132 @@
...
+            table = tabletask.output().table
+            column_names = [tabletask.id_to_column(f) for f in data[0]['fields']]
+            col_expression = '+'.join(['{table}.{column}'.format(table=table, column=cn) for cn in column_names])
+
+            return table, col_expression
+
+        in_colnames = []  # Column names in source tables
+        out_colnames = []  # Column names in destination table
+        from_tables = set()  # Set of all source tables / CTEs (FROM generation)
+        ctes = defaultdict(lambda: defaultdict(list))  # CTEs to be generated. {t1t2: { t1: [c1, c2], t2: [d1, d2]}}
```
ethervoid: ❤️

tasks/uk/census/wrapper.py
ethervoid: Au contraire https://www.python.org/dev/peps/pep-0020/ is non-pythonic. Cleverish is bad!

tasks/uk/census/scotland.py
javitonino: It's different per-task. The idea is to have a method to normalize column names, but the normalization is different for Scotland and UK. I will probably create an abstract class to make it more obvious.

tasks/uk/census/wrapper.py
```diff
@@ -0,0 +1,132 @@
...
+from tasks.util import MetaWrapper, TableTask
+from tasks.uk.cdrc import OutputAreas, OutputAreaColumns
+from tasks.uk.census.metadata import CensusColumns
+
+from .ons import ImportUK, ImportEnglandWalesLocal
+from .scotland import ImportScotland
+from .metadata import parse_table, COLUMNS_DEFINITION
+
+
+class Census(TableTask):
```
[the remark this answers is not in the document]
javitonino: Uhm, but I want a single `TableTask` for everything. The idea being, that you want data for the entire UK. i want to be able to put points in England, Scotland and NI, and ask for the total population of all of them at once. In order to do that, all the data from the three countries must be in the same table, thus, a single `TableTask`.

If you create three `TableTask`s, DO has no way to determine which table to use when augmenting a point, unless you differentiate by measure, timespan or geometry, which IMHO, does not make sense in this case.

tasks/uk/census/scotland.py
javitonino: I'm probably removing it, since there is a check later down the process (in `wrapper`). It will also allow us to use `COPY` instead of the multi-insert (which is slow, because it lies and doesn't batch).

tasks/uk/census/wrapper.py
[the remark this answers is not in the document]
antoniocarlon: Understood then, and with that point of view it's undoubtedly correct. Thanks for the explanation! :)

ethervoid:
After checking and thinking about the more complex part (the mega-sql generator) this is my two cents summary:

- I didn't reach a better/simpler solution to consolidate all the data in on point 
- Maybe we can incur in problems with the limit of the table name in the CTEs
- In case of performance problems we could write the query result to file and then use COPY
- We have to make a `complete` for that table. Maybe storing the number of rows we need to have and check with the inserted or something like that

javitonino:
> We have to make a complete for that table. Maybe storing the number of rows we need to have and check with the inserted or something like that

I don't follow you. The current method checks that the table exists and has rows. Why do you think we should check the number of rows?

ethervoid:
This is something I keep seeing. We check that we have a table and it has rows but maybe we missed some of them to be inserted. So, can we be more accurate to check if it has the right amount of rows?

javitonino:
Uhm, that feels more like a test to me. IMHO: `complete()`/`Target`s should be kept simple so resolution of dependencies can go fast. This kind of check which I agree can be useful, seems better suited for a test.

Maybe creating some kind of test suite (could be a task) for each import task, and wrapping it in the Makefile?

In any case, I would prefer to have this check as part of `run` rather than `complete()` (an `assert`). Luigi does not run the `complete()` method after finishing a task (only when this task is used as a dependency for other), so it would not get executed in this case.

ethervoid:
> Uhm, that feels more like a test to me. IMHO: complete()/Targets should be kept simple so resolution of dependencies can go fast

Agree but we're could just insert 1 row and be good with that we could do it with the test as you said but is something we have to check.

> In any case, I would prefer to have this check as part of run rather than complete() (an assert). Luigi does not run the complete() method after finishing a task (only when this task is used as a dependency for other)

Yes, and `Census` is a dependency for `CensusWrapper`

But in any case, we should come to a solution for the `TableTarget.exists` method to be better than just check table and it has rows. But just is a thought nothing more.

javitonino:
Some major changes:
- Created `/lib` directory to store code that can be reused. The idea is to put here things that are not `tasks`, so it's easier to differentiate. What do you think?
- No more shell commands, using COPY everywhere, Extracted `COPY` code with psycopg, so it can be reused
- Created `DirectoryTarget`. It is a wrapper of `luigi.LocalTarget` which handles the temporary directory creation/deletion better (i.e: automatic). I only use it in one task, because the rest inherit from `DownloadUnzipTask` and I don't feel like refactoring even more things (download code) in this PR
- In general, address all PR comments

antoniocarlon:
I love the `/lib` directory. In general anything with a better (more descriptive) name than `util` is great.

tasks/uk/census/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+def sanitize_identifier(colid):
+    return DISALLOWED_CHARACTERS_RE.sub('_', '_'.join(colid.split(';')[0].split(':')[-2:]))
+
+
+def parse_table(table_id):
+    '''
+    Returns a tuple like (KS, NI) from a table string like KS201NI
+    '''
+    return re.match('([A-Z]{2})\d+([A-Z]{2})[a-z]?', table_id).groups()
```
ethervoid: You have the same `compile` issue here. Just if you want to add as a constant as DISALLOWED_CHARACTERS_RE

tasks/uk/census/scotland.py
```diff
@@ -0,0 +1,52 @@
...
+
+            for c in header[1:]:
+                cols[self.id_to_column(c)] = 'NUMERIC'
+
+        with open(infile) as csvfile:
+            copy_from_csv(
+                current_session(),
+                self.output().table,
+                cols,
+                CSVNormalizerStream(csvfile, lambda row: ['0' if f == '-' else f.replace(',', '') for f in row])
```
ethervoid: ❤️

tasks/uk/census/metadata.py
javitonino: I could swear I had committed this. It seems I reverted it with one too many Ctrol+Z's. I'll commit it after you finish your review pass.

tasks/uk/census/ons.py
```diff
@@ -0,0 +1,126 @@
...
+    GEO_TYPES = [
+        'TYPE258',  # Small Areas (Northern Ireland)
+        'TYPE299'   # Output Areas (England, Wales, Scotland)
+    ]
+
+    table = Parameter()
+
+    def run(self):
+        # Query API, extract table ID from name
+        meta = requests.get(self.API_URL.format(self.table)).json()
```
ethervoid: :_)

ethervoid:
Done the review. Looks good to me. 

Outstanding work in this revamp. Especially love the utils class for CSV and COPY

lib/targets.py
```diff
@@ -0,0 +1,41 @@
...
+
+from tasks.util import classpath
+
+
+class DirectoryTarget(Target):
+    '''
+    Similar to :class:`LocalTarget <luigi.LocalTarget>` but automatically creates and destroys temp directories
+    '''
+    def __init__(self, task):
+        self.path = os.path.join('tmp', classpath(task), task.task_id)
```
antoniocarlon: As this class is `DirectoryTarget` and not `DODirectoryTarget` I would pass the `path` as the parameter instead of the `task`

tasks/uk/census/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class SourceTags(TagsTask):
```
antoniocarlon: As a reminder for a future refactor, we should also make the tags creation declarative

tasks/util.py
```diff
@@ -834,6 +834,9 @@ def version(self):
         '''
         return 0
 
+    def prefix(self):
```
antoniocarlon: :heart:

lib/targets.py
javitonino: This is a pattern I have seen in a lot of tasks, and I prefer discharging the task of this, since the classpath thing will be needed in many places.

I agree that passing the task here is not the cleanest though, so open to suggestions. I thought about doing a small mixin class for tasks. Something like:
```python
class FileSystemTask:
  def output_path(self):
    return os.path.join('tmp', classpath(self), self.task_id)

class DownloadUK(DownloadUnzipTask, FileSystemTask):
  def output():
    return DirectoryTarget(path=self.output_path())
```

But it seems more complicated and I don't know if it is very useful. That's why I went with the current approach, and we can refine it as we add more tasks using it.

antoniocarlon: It's not so horrible, I was just thinking out loud

tasks/uk/census/scotland.py
```diff
@@ -0,0 +1,52 @@
...
+from tasks.meta import current_session
+
+from .metadata import sanitize_identifier
+
+
+class DownloadScotlandLocal(DownloadUnzipTask):
+    URL = 'http://www.scotlandscensus.gov.uk/ods-web/download/getDownloadFile.html?downloadFileIds=Output%20Area%20blk'
+
+    def download(self):
+        urllib.retrieve(self.URL, '{}.zip'.format(self.output().path))
```
antoniocarlon: Did you mean `urllib.urlretrieve`?

javitonino: Uhm, yeah, sorry :(

PR #344, opened 2017-10-13
Py3

javitonino:
- Upgrade to Xenial + Python3 in `Dockerfile`
- Remove unused focker images
- Upgrade to Python3 (running `2to3`)
- Removed a crontab that was supposed to autodump images.
- [Autonamespace](https://github.com/CartoDB/bigmetadata/pull/344/files#diff-ff17943812ec2171c673c7b301d38bc8) tasks with the module. This makes possible the execution of different ETLs in parallel (before it broke due to naming conflicts in tasks). This affected the catalog checks if a PR affected more than one country.

Using a label for the image, so we can have the tests run in Travis :)

**Important**: Before deploying: revert https://github.com/CartoDB/bigmetadata/pull/344/commits/e6808ad91db1009557d962bbfc481578214c1b92 and upload docker image to latest tag.


docker-compose.yml
```diff
@@ -86,36 +79,5 @@
     "env_file": ".env",
     "stdin_open": true,
     "tty": true
-  },
-  "ipython": {
```
ethervoid: I think @michellemho uses iPython but I'm not sure

docker-compose.yml
```diff
@@ -86,36 +79,5 @@
...
-      "PGPASSWORD": "docker",
-      "PGDATABASE": "test",
-      "LC_ALL": "C.UTF-8",
-      "LANG": "C.UTF-8"
-    },
-    "env_file": ".env",
-    "working_dir": "/tmp/ipython",
-    "command": "sh -c /bigmetadata/ipython/start.sh"
-  },
-  "redis": {
```
ethervoid: Why do we have a redis image?

tasks/__init__.py
```diff
@@ -0,0 +1,3 @@
+import luigi
+
+luigi.auto_namespace()
```
ethervoid: ❤️

tasks/au/data.py
```diff
@@ -349,28 +349,28 @@ def requires(self):
         }
 
     def timespan(self):
-        return unicode(self.year)
+        return str(self.year)
```
ethervoid: ❤️  python3 unicode part much better than python2

docker-compose.yml
javitonino: No idea. It was linked to the luigi daemon, and the daemon supports using redis as a backend. However, the current configuration uses psql as the backend, so I guess it's just legacy.

tasks/br/data.py
```diff
@@ -231,7 +231,7 @@ def columns(self):
         filepath = "meta/{tablename}.csv".format(tablename=self.tablename)
 
         session = current_session()
-        with open(os.path.join(os.path.dirname(__file__),filepath)) as tsvfile:
+        with open(os.path.join(os.path.dirname(__file__), filepath), encoding='latin1') as tsvfile:
```
ethervoid: Is mandatory to include the encoding? In python2 the default is latin1 and now is utf-8 in p3?

docker-compose.yml
javitonino: If it's in use, we can keep it, but I'd prefer to have it removed or as an optional dependency, since it's quite a big image.

tasks/br/data.py
```diff
@@ -240,8 +240,7 @@ def columns(self):
                     continue
 
                 col_id = line[0]
-                col_name_pt = line[1].decode('latin1')
-                col_name_en = line[2]
+                col_name_pt = line[1]
```
ethervoid: Why do you remove the col_name_en column?

javitonino: flake8 warning about unused variable.

tasks/es/ine.py
```diff
@@ -86,7 +86,7 @@ def columns(self):
                   boundary_type['cartographic_boundary']],
         )
         cusec_id = OBSColumn(
-            name=u"Secci\xf3n Censal",
+            name="Secci\xf3n Censal",
```
ethervoid: 🤦‍♂️

tasks/br/data.py
javitonino: This is a bit bold, but I think for the better. In py2, this returned undecoded strings (byte arrays), while in py3, it decodes it to utf-8. If you see a bit below, there was a bit of code to decode as latin1 the name field. In py3, the best way to do this is to pass the encoding to use for the whole file, otherwise you get decoding errors while opening it.

tasks/es/ine.py
javitonino: We can probably change that to `Sección Censal` now, without having to use the escape characters.

tasks/util.py
```diff
@@ -1231,7 +1231,7 @@ def output(self):
         table lives in a special-purpose schema in Postgres derived using
         :func:`~.util.classpath`.
         '''
-        return PostgresTarget(classpath(self), self.task_id)
+        return PostgresTarget(classpath(self), self.task_id.split('.')[-1])
```
ethervoid: Just to be sure, this is now need because of the auto_namespace option activated isn't it?

docker-compose.yml
ethervoid: Ok, lets wait for Michelle answer

tasks/br/data.py
ethervoid: Ok, looks good to me too

tasks/br/data.py
ethervoid: 👍

tasks/es/ine.py
ethervoid: Yep, now with p3 we should after this change is merged

tasks/util.py
javitonino: Yes. It's not actually needed, but this makes the tables inside the schema keep the same name, which means we don't have to reimport them.

ethervoid:
### Acceptance

I've found some issues with es.ine and fr and more could come and should be solved when they arise. 

I'd merge this and fix the problems when they appear

PR #353, opened 2017-10-18
Add update date to boundaries. Added boundaries queries

### LINKED ISSUE #345: Add update date to boundaries in catalog
@javitonino

antoniocarlon:
- Added timespans to boundaries
- Added queries to update another table with the boundaries

Resolves #345

### DIFF
--- a/catalog/templates/column.html
+++ b/catalog/templates/column.html
@@ -54,7 +54,7 @@ Code example
       .. code-block:: postgresql
 
           WITH data AS (
-            SELECT cartodb_id AS id
+            SELECT geo_id AS id
             FROM <table_to_update>
           ),
           meta AS (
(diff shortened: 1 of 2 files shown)


catalog/templates/column.html
```diff
@@ -48,12 +49,34 @@
...
+                (SELECT meta FROM meta)) obs_data
+          WHERE <table_to_update>.cartodb_id = obs_data.id
+
       Obtain boundaries matched to another table's ID column.
 
       .. code-block:: postgresql
 
           WITH ids AS (
-          SELECT ARRAY_AGG(<id_col>) ids
+          SELECT ARRAY_AGG(cartodb_id) ids
```
javitonino: Uhm, this is actually not `cartodb_id`, but rather, the geoid of the geometry. With cartodb_ids, the query doesn't work, as it expects a string array, not an integer array.

catalog/templates/column.html
```diff
@@ -48,12 +49,34 @@
...
+            ) meta
+          )
+          UPDATE <table_to_update> SET
+                geomref = data->0->>'geomref',
+                the_geom = (data->0->>'value')::Geometry,
+                the_geom_webmercator = ST_Transform((data->0->>'value')::Geometry, 3857)
+          FROM OBS_GetData(
+                (SELECT Array_agg(id) FROM data),
+                (SELECT meta FROM meta)) obs_data
+          WHERE <table_to_update>.cartodb_id = obs_data.id
```
javitonino: This query doesn't work at all in my account. I think it is related to the other comment about ids, but there are more things involved.

I think we should pass the geoids to this functions, and it can add the geometries (geocoding) to the original table. It seems this query is also trying to update the geomref, which doesn't make sense if that's the one being passed into the function.

Anyway, I feel like I may be missing something here 🤔

catalog/templates/column.html
antoniocarlon: You are right, the column must be `geo_id` instead of `cartodb_id`

catalog/templates/column.html
antoniocarlon: The problem was in the id column name. It must be `geo_id` instead of `cartodb_id` as you mentiones in the other comment

javitonino:
Now it works 🙌 

Just one minor fix before deploying, can we put `geoid` inside `< >`? This field should be updated to reflect the user table structure.

PR #357, opened 2017-10-24
Create a Luigi task to simplify the geographies

### LINKED ISSUE #338: Find the best simplification mechanism/factor for each geometry table
Find the best simplification mechanism/factor for each geometry table
 - [x] Australia
 - [x] Brazil
 - [x] Canada
 - [x] Spain
 - [x] Eurostat
 - [x] France
 - [x] Mexico
 - [x] United Kingdom
 - [x] United States

antoniocarlon:
Closes #338

- Created a Luigi task that allows to simplify the geographies one by one or in batch
- Found the best simplification factor and mechanism for every table
- Benchmarked the simplifications


tasks/simplify.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class SimplifyOne(Task):
+    table = Parameter()
+
+    def run(self):
+        with open(os.path.join(os.path.dirname(__file__), 'simplifications.json')) as file:
+            tables = json.load(file)
+
+            if self.table in tables:
```
javitonino: I would remove this check, and let an exception be thrown if the table passed to this task is not defined in the JSON file. It is an error after all, and I think in this case it is preferable to fail than to silently eat the error.

tasks/simplify.py
```diff
@@ -0,0 +1,55 @@
...
+                table = tables.get(self.table)
+                LOGGER.info("Simplifying %s (%s) using %s", self.table, table.get(TABLE_ID), table.get(SIMPLIFICATION))
+
+                simplification = table.get(SIMPLIFICATION)
```
javitonino: In this cases, where we expect to be a value, I would replace all `.get` calls with array accesses `[]`. A missing parameter is an error here, is it not?

tasks/simplify.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class SimplifyAll(Task):
+    def requires(self):
+        with open(os.path.join(os.path.dirname(__file__), 'simplifications.json')) as file:
+            tables = json.load(file)
+
+            tasks = []
+            for table in tables:
+                tasks.append(SimplifyOne(table=table))
```
javitonino: You can just do a list comprehension!
```python
return [SimplifyOne(table=table) for table in json.load(file)]
```

tasks/simplifications.json
```diff
@@ -0,0 +1,550 @@
+{
+	"obs_b9cfbb7fddb54af1cfec091140d7ca8cc7aa753e": {
```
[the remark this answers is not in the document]
antoniocarlon: Yes, that sounds better and makes ir more reusable :+1:

tasks/simplify.py
antoniocarlon: :man_facepalming: very long time being a Java guy...

tasks/simplify.py
```diff
@@ -0,0 +1,55 @@
...
+def find_table_name(table_id):
+    session = CurrentSession().get()
+    return session.execute("SELECT tablename FROM observatory.obs_table WHERE id = '{tableid}'".format(
+                            tableid=table_id)).fetchone()[0]
+
+
+class SimplifyOne(Task):
+    table = Parameter()
+
+    def run(self):
```
javitonino: What about making this a wrapper task? I think it fits better, because this is not actually a dynamic dependency, the dependency can be known before runtime (it just depends on the static json and a parameter).

It would be a matter of:
- Inheriting from `WrapperTask`
- Putting the code in `requires` rather than `run`
- `return`ing instrad of `yield`ind the tasks.

tasks/simplify.py
```diff
@@ -0,0 +1,55 @@
...
+                                                  skipfailures=table[SKIP_FAILURES], maxmemory=table[MAX_MEMORY])
+            elif simplification == SIMPLIFICATION_POSTGIS:
+                yield SimplifyGeometriesPostGIS(schema=OBSERVATORY_SCHEMA, table_input=tablename,
+                                                geomfield=table[GEOM_FIELD], retainfactor=table[FACTOR])
+            else:
+                raise ValueError('Invalid simplification "{simplification}" for {table}'.format(
+                    simplification=simplification, table=self.table))
+
+
+class SimplifyAll(Task):
```
javitonino: This could also be made a WrapperTask

tasks/simplify.py
antoniocarlon: Makes a lot of sense as [the documentation](http://luigi.readthedocs.io/en/stable/luigi_patterns.html#triggering-many-tasks) states that:
> You’ll need to use `WrapperTask` for this instead of the usual `Task` class, because this job will not produce any output of its own, and as such needs a way to indicate when it’s complete. This class is used for tasks that only wrap other tasks and that by definition are done if all their requirements exist.

javitonino:
LGTM. However, tests are broken due to imports. From the looks of it, it's probably related with py3. I solved this kind of thing with absolute imports (`tasks.util`) or explicit relative imports (`.util`).

tasks/simplifications.json
```diff
@@ -209,21 +209,21 @@
 		"maxmemory": "8192",
 		"skipfailures": "no"
 	},
-	"eu.geo.nuts_geometries_0_5af288f97f": {
+	"eu.geo.tasks_eu_geo_nuts_geometries_0_5af288f97f": {
```
javitonino: Maybe we want to fix this in `TableTask` instead, so the ID doesn't change? We talked about this, but I only had the hashed table names in mind, maybe the decision is different for this kind of ids? also cc @ethervoid

antoniocarlon: I'm undoing this changes (related to https://github.com/CartoDB/bigmetadata/pull/366)

PR #358, opened 2017-10-25
Updated run and run-parallel Makefile tasks

antoniocarlon:
As now we need to completely qualify the Task name, there is no need to hide the `tasks` part in the `module` Luigi parameter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -286,178 +286,118 @@ deps-tree:
 
 ### au
 au-all:
-	docker-compose run --rm bigmetadata luigi \
-	  --module tasks.au.data tasks.au.data.BCPAllGeographiesAllTables --year 2011 \
-	  --parallel-scheduling --workers=8
+	 au.data.BCPAllGeographiesAllTables --year 2011
 
 au-geo:
-	docker-compose run --rm bigmetadata luigi \
-	  --module tasks.au.geo tasks.au.geo.AllGeographies --year 2011 \
-	  --parallel-scheduling --workers=8
+	make -- run-parallel au.geo.AllGeographies --year 2011
 
 ### br
 br-all: br-geo br-census
 
 br-census:
-	docker-compose run --rm bigmetadata luigi \
-		--module tasks.br.data tasks.br.data.CensosAllGeographiesAllTables \
-		--parallel-scheduling --workers=8
+	make -- run-parallel br.data.CensosAllGeographiesAllTables
 
 br-geo:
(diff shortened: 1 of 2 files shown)


javitonino:
An idea, because I'm lazy. Now that the task is fully qualified, we can calculate the `--module` part automatically (it's the task name except the part after the last dot). That way we could do `make run tasks.es.ine.PopulationMetaWrapper` (and avoid writing `tasks.es.ine` twice). What do you think?

antoniocarlon:
Avoid unnecesary typing? I'm in :+1:

Makefile
```diff
@@ -107,17 +107,29 @@ kill:
...
 ifeq (run,$(firstword $(MAKECMDGOALS)))
-  # use the rest as arguments for "run"
-  RUN_ARGS := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+  # https://stackoverflow.com/questions/10571658/gnu-make-convert-spaces-to-colons
+  space :=
+  space +=
+  # From word 2 to the end is the task
+  TASK := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
+  # Split the quelified task into words and remove the last word (task name) to get the module name
+  MOD_NAME := $(subst $(space),.,$(filter-out $(word $(words $(subst ., ,$(wordlist 2,2,$(MAKECMDGOALS)))),$(subst ., ,$(wordlist 2,2,$(MAKECMDGOALS)))), $(subst ., ,$(wordlist 2,2,$(MAKECMDGOALS)))))
```
javitonino: Wow! That Makefile syntax!

Trying to see it from the outside, I asked Google how we could run shell commands in here. You can do something like `MOD_NAME != command`

Maybe something like `echo $(TASK) | sed "s/^\(.*\)\..*$/\1/"`? Still not very readable, but at least is just a regex, not some custom `make` thing.

Makefile
```diff
@@ -278,178 +286,118 @@ deps-tree:
 
 ### au
 au-all:
-	docker-compose run --rm bigmetadata luigi \
-	  --module tasks.au.data tasks.au.data.BCPAllGeographiesAllTables --year 2011 \
-	  --parallel-scheduling --workers=8
+	 au.data.BCPAllGeographiesAllTables --year 2011
```
javitonino: Missing `make run-parallel`

antoniocarlon:
Shame on me, damn copy-paste!

PR #366, opened 2017-10-31
Fixed ids in obs table

antoniocarlon:
- Created method to get the name of the task from the `task_id`
 - Using this method to get the `id` for the `TableTask`s

### DIFF
--- a/tasks/util.py
+++ b/tasks/util.py
@@ -108,9 +108,9 @@ def classpath(obj):
     return classpath_ if classpath_ else 'tmp'
 
 
-def proper_task_id(task_id):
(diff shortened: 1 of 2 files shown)


tasks/util.py
```diff
@@ -108,6 +108,13 @@ def classpath(obj):
     return classpath_ if classpath_ else 'tmp'
 
 
+def proper_task_name(task_id):
```
javitonino: Nitpicking: I would rename this to something like `unqualified_task_id`. The current name makes me think it's going to return the task name, but this also includes parameters and a hash.

Issue #311, opened 2017-09-11
Run ANALYZE on source tables in OBSMetaToLocal


javitonino:
`OBSMetaToLocal` can be very slow if the wrong query plan is used. `ANALYZE` usually fixes the issue, so adding it to the task makes sense. We should do it for all source tables (`obs_column` and friends).

javitonino:
I run into this problem again today. Over 40 minutes to generate the catalog before I decided to cancel it.

PR #371, opened 2017-11-09
Solve universe target issues

### LINKED ISSUE #326: 'universe' target should not be an option for normalization

antoniocarlon:
- Restored `universe` target
 - Restored `universe` target for `median` and `average` aggregated measurements
 - Added a check to avoid having `median` and `average` aggregated measurements without `universe` target 
 - Fixing other minor issues

Fixes #326

**Note for the Acceptance:** BLS (US), Australia and Canada (NHS and census) had some `median` and `average` aggregated measurements without `universe` target


tasks/util.py
```diff
@@ -1665,6 +1669,23 @@ def check_null_columns(self):
...
+        session = current_session()
+        result = session.execute("SELECT c.id, c.aggregate, cc.reltype "
+                                 "FROM observatory.obs_table t "
+                                 "INNER JOIN observatory.obs_column_table ct ON t.id = ct.table_id "
+                                 "INNER JOIN observatory.obs_column c ON ct.column_id = c.id "
+                                 "FULL JOIN observatory.obs_column_to_column cc ON c.id = cc.source_id "
+                                 "WHERE t.tablename = '{table}' "
+                                 "AND c.aggregate IN ('average', 'median') "
+                                 "AND (reltype IS NULL OR reltype <> 'universe')".format(
+                                     table=self.output()._tablename)).fetchall()
```
javitonino: I think this query also returns an error if an average/median column has a denominator. Does this case make sense?

I think it does (dividing an average of a class over the average of the total population, e.g: average income of engineers / average income).

tests/test_util.py
```diff
@@ -139,19 +139,21 @@ def test_column_target_many_inits():
...
 @with_setup(setup, teardown)
 def test_table_target_many_inits():
-    pop_col = ColumnTarget(OBSColumn(
-        id='tests.population',
-        type='Numeric',
-        name="Total Population",
-        description='The total number of all',
-        aggregate='sum',
-        weight=10), FakeTask())
+    total_pop = OBSColumn(
```
javitonino: Now that I see test changes, why don't we try to add a test for this?

tasks/util.py
```diff
@@ -1665,6 +1669,23 @@ def check_null_columns(self):
             raise ValueError('The following columns of the table "{table}" contain only NULL values: {columns}'.format(
                 table=self.output().table, columns=', '.join([x[0] for x in result])))
 
+    def check_universe_in_aggregations(self):
```
javitonino: Now that I see test changes, why don't we try to add a test for this check?

antoniocarlon: Added test forcing the `ValueError` to raise when there is a `median` or an `average` aggregation without `universe` denominator

tasks/util.py
antoniocarlon: Currently, there isn't any case like this in the ETL but it makes sense, having a median/average with both a universe and a denominator shouldn's trigger the check.

antoniocarlon:
1. This PR should fix all the median and average aggregated measurements without `universe` target. The note for the acceptance tries to help about what to test
2. There's no need for changes in the extension as I have updated the `OBS_Meta` generation (also, see the comments [here](https://github.com/CartoDB/bigmetadata/issues/326))

tasks/carto.py
```diff
@@ -490,6 +500,10 @@ class OBSMeta(Task):
         AND numer_ctag.column_id = numer_c.id
         AND numer_ctag.tag_id = numer_tag.id
         AND numer_c.id = leftjoined_denoms.all_numer_id
+        AND NOT EXISTS (
+          SELECT 1 FROM agg_wo_universe u
+          WHERE u.id = numer_c.id
+        )
```
antoniocarlon: I have also tested the possible loss of performance of adding this check and it's insignificant

javitonino: Nice to see. Not sure if we want to keep this in this query as it is redundant with the tasks check. My concern is not performance but readability, my head almost exploded reading this. If you prefer to keep it, let's add some comments, at least.

tasks/carto.py
```diff
@@ -490,6 +500,10 @@ class OBSMeta(Task):
         AND numer_ctag.column_id = numer_c.id
         AND numer_ctag.tag_id = numer_tag.id
         AND numer_c.id = leftjoined_denoms.all_numer_id
+        AND NOT EXISTS (
```
javitonino: We may want to add a check to remove the non-denominated versions of the universe numerators:

`AND NOT (numer_agg IN ('median', 'average') AND denom_id IS NULL)` or something along those lines.

**Edit**: There are no NULL denom_id entries in the `obs_meta` table (except if a column has no denominators at all). This idea doesn't work.

tasks/util.py
```diff
@@ -1665,6 +1669,24 @@ def check_null_columns(self):
...
+        session = current_session()
+        result = session.execute("SELECT c.id, c.aggregate, STRING_AGG(COALESCE(cc.reltype,''), ',') reltype "
+                                 "FROM observatory.obs_table t "
+                                 "INNER JOIN observatory.obs_column_table ct ON t.id = ct.table_id "
+                                 "INNER JOIN observatory.obs_column c ON ct.column_id = c.id "
+                                 "FULL JOIN observatory.obs_column_to_column cc ON c.id = cc.source_id "
+                                 "WHERE t.tablename = '{table}' "
+                                 "AND c.aggregate IN ('average', 'median') "
+                                 "GROUP BY 1, 2 "
+                                 "HAVING LOWER(STRING_AGG(COALESCE(cc.reltype,''), ',')) NOT LIKE '%universe%'".format(
```
javitonino: You could make this prettier with `ARRAY_AGG` and `ANY`

javitonino:
Like a charm, merging.

Small issues:
- au.data is very slow generating columns due to the yields. It seems a return+list comprehension is faster
- I manually reduced the version for some columns in order to reimport, since this was missing some obscure dumps (eg: ACS quantiles need to be bumped in order to be able to easily run the normal columns).

Al so, I run into the slow meta generation again, fixed it with an analyze. See https://github.com/CartoDB/bigmetadata/issues/311

PR #373, opened 2017-11-10
util.py refactor

antoniocarlon:
Refactor of `tasks/util.py` that tries to avoid the circular imports problem found when trying to add new functionality or modify an existing one


javitonino:
I took a look at this and my head exploded xD It looks fine, although is hard to tell. Only thing that bother me a little is the `tasks.tasks` name, which I find a bit confusing. I'd have chosen something like `common_tasks` to better separate this. However, if it's any effort to change, I wouldn't bother, we will continue the refactor bit by bit anyway.

As a bonus, nice cleanup of unused parameters!

antoniocarlon:
The refactor just tries to split `utils.py` into tasks, targets and utils but having to revisit a huge amount of files leaves the PR with a lot of changes.

I would change the `tasks.py` name to `base_tasks.py` instead of `common_tasks.py` if you agree.

javitonino:
Nice, let's give it a test!

javitonino:
I tested a full ETL import of ES + UK + meta and catalog generation. Everything worked like a charm 👏

PR #380, opened 2017-11-17
Add simplifications to the ETL pipeline

### LINKED ISSUE #372: Add the simplification task to the ETL pipeline
Currently we have manually executed the simplificaton for all the `OBS_XXX` tables. We need to embed the simplification task into the ETL pipeline

 - [x] Australia
 - [x] Brazil
 - [x] Canada
 - [x] Spain
 - [x] Eurostat
 - [x] France
 - [x] Mexico
 - [x] United Kingdom
 - [x] United States

antoniocarlon:
Closes #372

Embeds the simplifications into the ETL pipeline

Tested one table for each ETL task

### DIFF
--- a/tasks/au/data.py
+++ b/tasks/au/data.py
@@ -337,7 +337,7 @@ class BCP(TableTask):
     resolution = Parameter()
 
     def version(self):
-        return 2
+        return 3
 
     def requires(self):
         import_data = {}
--- a/tasks/au/geo.py
+++ b/tasks/au/geo.py
@@ -1,6 +1,7 @@
 from luigi import Parameter, WrapperTask
 
-from tasks.base_tasks import ColumnsTask, DownloadUnzipTask, Shp2TempTableTask, TableTask, TagsTask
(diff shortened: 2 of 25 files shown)


antoniocarlon:
Yes, I'm updating it

javitonino:
I started doing acceptance for this and worked like a charm for normal cases. However, it failed for `tasks.eu.geo.SimplifiedImportSHNNames__99914b932b` task. I guess it has to do with the inheritance of the `get_table_id` function (the issue is that it cannot find the simplification.json entry).

As a note, that I forgot to mention in the first CR round. I don't like the `yield` in the simplify task, but it is not a big problem here since it's only one. Maybe could be resolved by changing the inheritance a little (so `Simplify` inherits from `TempTableTask`) but it loses some flexibility. To make it clear, I'm ok on how it works now, although I prefer this to be handled through static requirements (since there is nothing dynamic in this). But if it's too complicated, don't bother, let's ship this.

antoniocarlon:
I'll take a look at it having your concerns in mind :)

PR #381, opened 2017-11-17
Fixed timespan naming

### LINKED ISSUE #379: Catalog example queries have the suggested name wrong
Related #376 

Right now we're composing the suggested_name of the column by hand using the `col.suggested_name` and adding the `col.timespan` as you can see [here](https://github.com/CartoDB/bigmetadata/pull/376/files#diff-d70579600c06a04ff503924c40445bb4R359), but this is creating wrong column names that don't match with the ones are returned by `OBS_GetMeta`.

Example:

[here](https://cartodb.github.io/bigmetadata/united_states/education/us.census.acs.B01003001.html) we're returning `total_pop_per_sq_km_2015 - 2015` but we should return `total_pop_per_sq_km_2015_2015`

//cc @michellemho

antoniocarlon:
Fixed timespan column naming in catalog

Fixes #379

### DIFF
--- a/tasks/sphinx.py
+++ b/tasks/sphinx.py
@@ -204,7 +204,7 @@ class GenerateRST(Task):
                     FIRST(c.aggregate),
                     JSONB_Object_Agg(t.type || '/' || t.id, t.name),
                     'name' suggested_name,
-                    STRING_AGG(DISTINCT tab.timespan, ',') timespan,
+                    ARRAY_AGG(DISTINCT tab.timespan) timespan,
                     ARRAY[]::Text[] denoms,
                     ARRAY[]::Text[],
                     ST_AsText(ST_Envelope(FIRST(tab.the_geom))) envelope
@@ -258,7 +258,7 @@ class GenerateRST(Task):
                     numer_aggregate,
                     numer_tags,
                     numer_colname suggested_name,
-                    numer_timespan timespan,
+                    ARRAY_AGG(numer_timespan) timespan,
                     ARRAY_AGG(DISTINCT ARRAY[
                     denom_reltype,
                     denom_id,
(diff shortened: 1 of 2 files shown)


tasks/sphinx.py
```diff
@@ -204,7 +204,7 @@ def _boundaries_tree(self, section_id, subsection_id):
                     FIRST(c.aggregate),
                     JSONB_Object_Agg(t.type || '/' || t.id, t.name),
                     'name' suggested_name,
-                    ARRAY_AGG(DISTINCT tab.timespan) timespan,
+                    STRING_AGG(DISTINCT tab.timespan, ',') timespan,
```
javitonino: Uhm, what happens when there are multiple timespans? This + slugify will probably do something ugly for US where there are multiple timespans, no? (I'm imagining `2015_2010_2015_2015_2010_2010` or something along those lines)

antoniocarlon: Let me take a look

tasks/sphinx.py
```diff
@@ -300,6 +300,7 @@ def _parse_columns(self, all_columns_result):
                 'tags': col[6],
                 'suggested_name': col[7],
                 'timespan': col[8],
+                'timespan_sluggified': underscore_slugify('_'.join(col[8])),
```
javitonino: Doesn't this have the same problem than before? This will return all timespans joined by `_`, won't it? I would just do `col[8][0]` to return the default timespan.

antoniocarlon: We are not using `timespan_sluggified` for boundaries, just for numerators, but we are using the `_parse_columns` method for both and, as the boundaries query is returning an `ARRAY_AGG` for timespans, we also need to return an `ARRAY_AGG` for timespans for the numerators and manage both timespans as if they were arrays to avoid the forementioned problems.

Anyway you are right that `col[8][0]` would have the same effect but I think that getting an `ARRAY_AGG` and then keeping only the first element is more confusing than this.

PR #384, opened 2017-11-22
Solving errors executing ETL tasks (embedded simplification)

### LINKED ISSUE #383: Errors executing ETL tasks
The execution of the BR task with the simplification embedded has raised some errors that we need to solve

- CA ETL task:
The name `ca.statcan.geo.importgeography_pr_0` in the `simplifications.json` is not correct (should be `ca.statcan.geo.importgeography_pr_`)

- BR ETL task:
```
ValueError: The following columns of the table "observatory.obs_d8684e697defcba33e8e987ba58a010d0bdf8a0b" are aggregated as 'median' or 'average' but lack of 'universe' target: br.data.Basico_V011, br.data.Basico_V009, br.data.Basico_V005, br.data.Basico_V003, br.data.Basico_V007
```

- EU ETL task:
```
luigi.target.FileAlreadyExists: Destination exists: tmp/simplification/eu.geo.ImportSHNGeoms__99914b932b_simpl
```

- EU ETL task:
```
ERROR: [pid 1] Worker Worker(salt=124385866, workers=1, host=273161c668b1, username=root, pid=1) failed    tasks.eu.eurostat.TableEU(table_name=EDUC_RENRLRG1, subsection=education, nuts_level=2, unit=people, year=2003)

ValueError: The following columns of the table "observatory.obs_f4fcec0d07615826962078191fab954c9d3ebc19" contain only NULL values: educ_renrlrg1_f_nr_ed3_g, educ_renrlrg1_m_nr_ed3_g, educ_renrlrg1_t_nr_ed3_g, educ_renrlrg1_f_nr_ed3_vpv, educ_renrlrg1_m_nr_ed3_vpv, educ_renrlrg1_t_nr_ed3_vpv, educ_renrlrg1_f_nr_ed4_g, educ_renrlrg1_m_nr_ed4_g, educ_renrlrg1_t_nr_ed4_g, educ_renrlrg1_f_nr_ed4_vpv, educ_renrlrg1_m_nr_ed4_vpv, educ_renrlrg1_t_nr_ed4_vpv
```

- US ETL task: 
Can't find an entry in `simplifications.json` for some tables (`congressional_district` , 
`school_district_unified`, `census_tract`, `block_group`, `school_district_secondary`, `school_district_elementary`)

antoniocarlon:
- Solved CA json problem
 - Solved BR averages without universe denominator problem
 - Solved EU null columns problem + `luigi.target.FileAlreadyExists`
 - Solved US no simplification found for some tables

Fixes #383

### DIFF
--- a/tasks/br/data.py
+++ b/tasks/br/data.py
@@ -254,11 +254,11 @@ class Columns(ColumnsTask):
                 denominators = denominators.split('|')
 
                 targets_dict = {}
-                for x in denominators:
-                    x = x.strip()
+                for denom in denominators:
+                    denom = denom.strip()
                     target_type = 'universe' if col_agg in ['median', 'average'] else 'denominator'
-                    targets_dict[cols.get(x, column_reqs[x].get(session)
-                                 if x in column_reqs else None)] = target_type
+                    targets_dict[cols.get(denom, column_reqs[denom].get(session)
+                                 if denom in column_reqs else None)] = target_type
 
                 targets_dict.pop(None, None)
 
--- a/tasks/br/data.py
+++ b/tasks/br/data.py
@@ -202,7 +202,6 @@ class Columns(ColumnsTask):
 
         return True
 
-
     def columns(self):
         cols = OrderedDict()
         input_ = self.input()
@@ -257,7 +256,9 @@ class Columns(ColumnsTask):
                 targets_dict = {}
                 for x in denominators:
                     x = x.strip()
-                    targets_dict[cols.get(x, column_reqs[x].get(session) if x in column_reqs else None)] = 'denominator'
+                    target_type = 'universe' if col_agg in ['median', 'average'] else 'denominator'
+                    targets_dict[cols.get(x, column_reqs[x].get(session)
+                                 if x in column_reqs else None)] = target_type
 
                 targets_dict.pop(None, None)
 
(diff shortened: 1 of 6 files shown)


tasks/br/data.py
```diff
@@ -257,7 +256,9 @@ def columns(self):
                 targets_dict = {}
                 for x in denominators:
                     x = x.strip()
-                    targets_dict[cols.get(x, column_reqs[x].get(session) if x in column_reqs else None)] = 'denominator'
+                    target_type = 'universe' if col_agg in ['median', 'average'] else 'denominator'
+                    targets_dict[cols.get(x, column_reqs[x].get(session)
```
ethervoid: For readability, we should change `x` for `denom` or `denominator`

Another question, what is the effect to have a target with universe?

tasks/eu/geo.py
```diff
@@ -56,6 +56,7 @@ def output(self):
 
 
 class ImportSHNGeoms(Shp2TempTableTask):
+    level = IntParameter()
```
ethervoid: Why is this change needed?

tasks/simplifications.json
```diff
@@ -437,70 +437,70 @@
 		"geomfield": "the_geom",
 		"factor": "8",
 		"maxmemory": "8192",
-		"skipfailures": "no"
+		"skipfailures": "yes"
```
ethervoid: Why do we have to enable this?

antoniocarlon: The TIGER geometries are the reason to have the `skipfailures` property in this JSON in the first place. `ogr2ogr` gives a lot of warnings but returns an error signal that makes the task fail.

tasks/eu/geo.py
antoniocarlon: In EU we have three different simplifications for the three existing levels (see [here](https://github.com/CartoDB/bigmetadata/blob/08e5b5fafe65f0469bc108c83b0e34ac4d76eab9/tasks/simplifications.json#L248-L271)). We need a to identify the level with the task id

tasks/br/data.py
antoniocarlon: universes are used to denormalize average and median aggregations (it doesn't make sense to normalize an average for example using the area of the geometry).

It's used [here](https://github.com/CartoDB/observatory-extension/blob/6066ef028dc428d409baa6e64bf2010fa0f38064/src/pg/sql/41_observatory_augmentation.sql#L708-L724) and [here](https://github.com/CartoDB/bigmetadata/issues/326) you can find an extended explanation

tasks/simplifications.json
ethervoid: Yep but we have it disabled and now is enable. Why didn't we have it enabled and why is required now?

antoniocarlon: We manually simplified the `obs_xxx` tables for TIGER (final ETL tables) and now we are automatically simplifying the original TIGER tables

PR #386, opened 2017-11-24
Create obs table to table

### LINKED ISSUE #385: Create OBS_Table_To_Table
Create `OBS_Table_To_Table` as the relation between two `OBS_Table`s (numerator to geometry)

antoniocarlon:
Creates the `observatory.OBS_Table_To_Table` table. To use it in a `TableTask`:
```
def targets(self):
    return {self.input()['geo'].obs_table: GEOM_REF}
```
Closes #385

### DIFF
--- a/tasks/meta.py
+++ b/tasks/meta.py
@@ -6,23 +6,17 @@ functions for persisting metadata about tables loaded via ETL
 
 import os
 import re
-import weakref
 
-import luigi
-from luigi import Task, BoolParameter, Target, Event
+from luigi import Target
 
-from sqlalchemy import (Column, Integer, Text, Boolean, MetaData, Numeric, cast,
-                        create_engine, event, ForeignKey, PrimaryKeyConstraint,
-                        ForeignKeyConstraint, Table, exc, func, UniqueConstraint,
-                        Index)
+from sqlalchemy import (Column, Integer, Text, MetaData, Numeric, cast,
+                        create_engine, event, ForeignKey, ForeignKeyConstraint,
+                        exc, func, UniqueConstraint, Index)
 from sqlalchemy.dialects.postgresql import JSON
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import relationship, sessionmaker, composite, backref
-from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm import relationship, sessionmaker, backref
 from sqlalchemy.orm.collections import attribute_mapped_collection
-from sqlalchemy.schema import ForeignKeyConstraint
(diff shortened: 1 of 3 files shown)


tasks/meta.py
```diff
@@ -6,23 +6,17 @@
...
-                        create_engine, event, ForeignKey, PrimaryKeyConstraint,
-                        ForeignKeyConstraint, Table, exc, func, UniqueConstraint,
-                        Index)
+from sqlalchemy import (Column, Integer, Text, MetaData, Numeric, cast,
+                        create_engine, event, ForeignKey, ForeignKeyConstraint,
+                        exc, func, UniqueConstraint, Index)
 from sqlalchemy.dialects.postgresql import JSON
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import relationship, sessionmaker, composite, backref
```
ethervoid: I assume this imports are not longer required

tasks/meta.py
```diff
@@ -269,6 +263,10 @@ def targets_creator(coltarget_or_col, reltype):
     return OBSColumnToColumn(target=col, reltype=reltype)
 
 
+def tabletargets_creator(target, reltype):
```
ethervoid: I'd name it `tabletotabletargets_creator`

tasks/meta.py
```diff
@@ -254,7 +248,7 @@ def tag_creator(tagtarget):
     return coltag
 
 
-def targets_creator(coltarget_or_col, reltype):
+def columntargets_creator(coltarget_or_col, reltype):
```
ethervoid: And here I'd name it coltocoltargets_creator or something like that

tasks/targets.py
```diff
@@ -219,10 +219,10 @@ def __init__(self, schema, name, obs_table, columns, task):
         obs_table.tablename = '{prefix}{name}'.format(prefix=OBSERVATORY_PREFIX, name=sha1(
             underscore_slugify(self._id).encode('utf-8')).hexdigest())
         self.table = '{schema}.{table}'.format(schema=OBSERVATORY_SCHEMA, table=obs_table.tablename)
+        self.obs_table = obs_table
```
ethervoid: Any special reason to change the name?

antoniocarlon: As we are going to expose it (see [here](https://github.com/CartoDB/bigmetadata/pull/386)) I didn't want to follow the private variables naming convention

tasks/meta.py
antoniocarlon: We need `relationship`, `sessionmaker` and `backref` ([line 18](https://github.com/CartoDB/bigmetadata/pull/386/files#diff-ec4044a9a28b8936802647b9b5977c8dR18))

PR #389, opened 2017-11-27
Fix universe numerators

### LINKED ISSUE #387: Fix (remove?) numerators with denominators in different tables
Currently we have several numerators with denominators in different tables and the tests (an queries to OBS_GetData) are failing.

We need to fix (use different `UNIVERSE` columns) or remove this columns until `OBS_GetData` is fixed (https://github.com/CartoDB/observatory-extension/issues/320)

antoniocarlon:
- CA: Changed `universe`s
 - BR: Changed `universe`s
 - AU: Removed `universe`s

Fixes #387

### DIFF
--- a/tasks/br/meta/Basico.csv
+++ b/tasks/br/meta/Basico.csv
@@ -1,33 +1,33 @@
-NOME DA VARIçVEL,DESCRI‚ÌO DA VARIçVEL,DESCRIPTION OF THE VARIABLE,unit,denominators,subsection,agg
-Cod_setor,C—digo do setor,Industry code,,,,
-Cod das Grandes Regi›es,C—digo das Grandes Regi›es (Regi›es Geogr‡ficas),Code of Major Regions (Geographical Regions),,,,
-Nome das Grandes Regi›es,Nome das Grandes Regi›es (Regi›es Geogr‡ficas),Name of Major Regions (Geographical Regions),,,,
-Cod_UF,C—digo da Unidade da Federa‹o,Federation Unit Code,,,,
-Nome_da_UF,Nome da Unidade da Federa‹o,Name of the Federation Unit,,,,
-Cod_meso,C—digo da mesorregi‹o,Meso-region code,,,,
-Nome_da_meso,Nome da mesorregi‹o,Name of mesoregion,,,,
-Cod_micro,C—digo da microrregi‹o,Microregion code,,,,
-Nome_da_micro,Nome da microrregi‹o,Name of microregion,,,,
-Cod_RM,C—digo da regi‹o metropolitana ou RIDE,Code of the metropolitan region or RIDE,,,,
-Nome_da_RM,Nome da regi‹o metropolitana ou RIDE,Name of the metropolitan region or RIDE,,,,
-Cod_municipio,C—digo do munic’pio,Town code,,,,
-Nome_do_municipio,Nome do munic’pio,Name of municipality,,,,
-Cod_distrito,C—digo do distrito,District code,,,,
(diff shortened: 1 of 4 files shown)


tasks/br/meta/Basico.csv
```diff
@@ -1,33 +1,33 @@
...
-V001,Domic�lios particulares permanentes ou pessoas respons�veis por domic�lios particulares permanentes,Permanent private households or persons responsible for permanent private households,people,Pessoa13_V003,housing,sum
-V002,Moradores em domic�lios particulares permanentes ou popula��o residente em domic�lios particulares permanentes,Residents in permanent private households or population residing in permanent private households,people,Pessoa13_V002,housing,sum
-V003,M�dia do n�mero de moradores em domic�lios particulares permanentes (obtida pela divis�o de Var2 por Var1),Average number of residents in permanent private households (obtained by dividing Var2 by Var1),people,Basico_V001,housing,average
-V004,Vari�ncia do n�mero de moradores em domic�lios particulares permanentes,Variance of the number of residents in permanent private households,people,Basico_V002,housing,variance
-V005,Valor do rendimento nominal m�dio mensal das pessoas respons�veis por domic�lios particulares permanentes (com e sem rendimento),Nominal average monthly income of persons responsible for permanent private households (with and without income),money,ResponsavelRenda_V022,income|housing,average
```
ethervoid: Shouldn't we remove these columns instead of change the denominator? Or we have a valid denominator for these aggregations that are in the same table?

tasks/ca/statcan/cols_nhs.py
```diff
@@ -22955,7 +22955,7 @@ def columns(self):
             weight=3,
             aggregate='average',
             tags=[ca, unit_people, subsections['employment']],
-            targets={t004c001_t: UNIVERSE},)
```
ethervoid: Same question as above

tasks/simplifications.json
```diff
@@ -183,7 +183,7 @@
 		"maxmemory": "8192",
 		"skipfailures": "no"
 	},
-	"ca.statcan.geo.importgeography_pr_0": {
```
ethervoid: Why is this change in this PR? Shouldn't be [here](https://github.com/CartoDB/bigmetadata/issues/383)?

tasks/br/meta/Basico.csv
antoniocarlon: We use the universes to denormalize medians and averages but there is no "correct" field for it (we can just figure it out). We used a universe that was in other table so I changed it for a "valid enough" universe from the same table as the numerator, but I'm open to dicussion.

tasks/ca/statcan/cols_nhs.py
antoniocarlon: Same answer as above

tasks/br/meta/Basico.csv
ethervoid: I see, is what we talked about yesterday.

What would be the impact of removing these columns? Could explain "valid enough" in depth please?

antoniocarlon: If we remove the columns we won't be able to use them in DO analysis.

The old universes relations were the ones that I considered that could "best" normalize the numerators (see [here](https://github.com/CartoDB/bigmetadata/pull/371)).

The new universes relations are the ones that, given that at this moment we cannot use the old universes because they belong to a different table, I considered that could "best" normalize the numerators.

ethervoid: Yep I know we would lose those columns in the DO analysis, but I don't know if there are relevant ones until we fix the issue and we can use denominator from other table but as you have investigated better substitutes looks good to me

PR #412, opened 2017-12-19
Added OBSTimespan entity

### LINKED ISSUE #411: Create OBSTimespan entity
- Create `OBSTimespan` in `meta.py` that creates a table named `observatory.OBS_Timespan`
 - `observatory.OBS_Timespan` must have `from` and `to` `daterange` columns

antoniocarlon:
- Added `OBSTimespan` entity
- Changed all the `TableTask`s to use the new entity
- ~Created a file to hold the know existing timespans~
- Kept the existing `timespan` column in `OBS_Table` with the existing values to ensure backwards compatibility

Closes #411


antoniocarlon:
**Notes for the acceptance:**
 - Check several tasks and make sure that the `timespan` field for every `OBS_Table` has not changed (to ensure backwards compatibility)
 - Check that every `OBS_Table` has a `timespan` related to an existing `OBS_Timespan` entry

ethervoid:
Before starting doing the CR I had some doubts about the implementation using a JSON file instead of auto parse the timespan strings. 

From my point of view, we could achieve the same result letting every task to parse it's own timespans and storing them in a format we define. There are two main points to have a JSON file, as @antoniocarlon said to me:
  - One is to be forced to introduce the new timespan so we can check if we already have this timespan and reuse it but this could be done using an alias filed storing all the related timespan.
  - Other is to have backward compatibility storing the old timespan name as ID

But we can do it automatically in every task achieving the same result and we could store them using our own known timespan format:
  - If we have our own format we don't have to be worried to check for existing timespans so we can parse and store them automatically
  - To maintain backward compatibility we can add an `alias` column that could be checked so we can store there the "old" values for that timespan and at some point we can get rid of it

This is an ongoing discussion and @antoniocarlon is thinking about why is the JSON file required.

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
+import re
```
ethervoid: Looks like a good first implementation. I'd have created a base abstract class and then moved the `parse_timespan` logic to the ETL classes.

These time spans are pretty common but in the future, we could end up having some tightly coupled ones but that's a "future"

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
...
+    return current_session().merge(timespan)
+
+
+def parse_timespan(timespan_id):
+    timespan_str = str(timespan_id).replace(' ', '')
+
+    ts_id, ts_name, ts_description, ts_timespan = None, None, None, None
+
+    # Year, for example '2017'
+    if re.match('\d{4}$', timespan_str):
```
ethervoid: Due to we have a "closed" regexp comparison, I'd use `^\d{4}$` regexp clause to avoid parse things like `pepito2004` (all the regexp we have should start with ^)

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
...
+        ts_timespan = '[{year}-{month:02d}-01, {year}-{month:02d}-{last_day}]'.format(year=year, month=month,
+                                                                                      last_day=last_day_of_month)
+    # Year and quarter, for example '2017Q3'
+    elif re.match('\d{4}Q\d{1}$', timespan_str):
+        year, quarter = list(map(int, timespan_str.split('Q')))
+        if quarter > 4:
+            raise ValueError("Invalid timespan '{timespan}' (invalid quarter)".format(timespan=timespan_id))
+
+        last_month = quarter * 3
+        initial_mont = last_month - 2
```
ethervoid: Typo s/initial_mont/initial_month/

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
...
+        ts_id = '{year}Q{quarter}'.format(year=year, quarter=quarter)
+        ts_name = '{year}Q{quarter}'.format(year=year, quarter=quarter)
+        ts_description = '{year}, {quarter} quarter'.format(year=year, quarter=QUARTERS[quarter])
+        ts_timespan = '[{year}-{ini_month:02d}-01, {year}-{end_month:02d}-{last_day}]'.format(year=year,
+                                                                                              ini_month=initial_mont,
+                                                                                              end_month=last_month,
+                                                                                              last_day=last_day)
+    # Day, month, year, for example '20171231'
+    elif re.match('\d{8}$', timespan_str):
+        year = timespan_str[:4]
```
ethervoid: I'd use `from dateutil.parser import parse` and then `parse(timespan_str)

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
...
+        month = timespan_str[4:6]
+        day = timespan_str[-2:]
+
+        if int(month) > 12:
```
ethervoid: Same as before, check for min and max bound

ethervoid: Using what I put above you don't need to cast to `int`

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError("Invalid timespan '{timespan}' (invalid month)".format(timespan=timespan_id))
+
+        last_day_of_month = calendar.monthrange(int(year), int(month))[1]
+        if int(day) > last_day_of_month:
+            raise ValueError("Invalid timespan '{timespan}' (invalid day)".format(timespan=timespan_id))
+
+        ts_id = '{year}{month}{day}'.format(year=year, month=month, day=day)
```
ethervoid: Using the `datetime` object returned by the `parse` object, you could use 'd.strftime('%Y%m%d')` to get the final attributes using the desired format

lib/timespan.py
```diff
@@ -0,0 +1,89 @@
...
+QUARTERS = {1: 'first',
+            2: 'second',
+            3: 'third',
+            4: 'fourth'}
+
+
+def get_timespan(timespan_id):
+    ts_id, ts_name, ts_description, ts_timespan = parse_timespan(timespan_id)
+    timespan = OBSTimespan(id=ts_id, name=ts_name, description=ts_description, timespan=ts_timespan)
+    return current_session().merge(timespan)
```
ethervoid: ph34r with adding objects to the sessions but is the way we're currently doing it

lib/timespan.py
antoniocarlon: Not needed if we use `parse(timespan_str)`

antoniocarlon: Not needed if we use `parse(timespan_str)`

tasks/us/bls.py
```diff
@@ -221,8 +222,8 @@ def requires(self):
                     )
         return requirements
 
-    def timespan(self):
-        return '{year}Q{qtr}'.format(year=self.year, qtr=self.qtr)
+    def table_timespan(self):
+        return get_timespan('{year}Q{qtr}'.format(year=self.year, qtr=self.qtr))
```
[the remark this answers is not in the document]
antoniocarlon: You're right, I have created the `alias`. I have left the `id` as is, but having the `alias` column we can change it whenever we want

lib/timespan.py
ethervoid: What about this?

ethervoid:
### Acceptance

- [x] Spain
- [x] Brasil
- [ ] Australia

ethervoid:
### Eurostat failing 
```
ERROR: [pid 1] Worker Worker(salt=900995649, workers=1, host=deceeebc7ab2, username=root, pid=1) failed    tasks.eu.eurostat.DownloadEurostat(table_code=NAMA_AUX_CRA)^M
Traceback (most recent call last):^M
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 194, in run^M
    new_deps = self._run_get_new_deps()^M
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 131, in _run_get_new_deps^M
    task_gen = self.task.run()^M
  File "/bigmetadata/tasks/eu/eurostat.py", line 39, in run^M
    shell('gunzip {output}.gz'.format(output=self.output().path))^M
  File "/bigmetadata/tasks/util.py", line 31, in shell^M
    return subprocess.check_output(cmd, shell=True, stderr=subprocess.STDOUT).decode(encoding)^M
  File "/usr/lib/python3.5/subprocess.py", line 626, in check_output^M
    **kwargs).stdout^M
  File "/usr/lib/python3.5/subprocess.py", line 708, in run^M
    output=stdout, stderr=stderr)^M
subprocess.CalledProcessError: Command 'gunzip tmp/eu.eurostat/tasks.eu.eurostat.DownloadEurostat_NAMA_AUX_CRA_241906624b.gz' returned non-zero exit status 1
```

ethervoid:
Blocked until #426 is solved

ethervoid:
Blocked until #433 is solved

ethervoid:
Blocked du to #437

PR #431, opened 2018-02-20
Fallback to unzip for deflate64 format

juanignaciosl:
Fixes #427.

Issue #427, opened 2018-02-19
Uncompressing Canada 2011 census fails


juanignaciosl:
It looks like the file has changed with an incompatible format:

```
✗ make -- run ca.statcan.data.DownloadData --resolution da_ --survey census
docker-compose run --rm bigmetadata luigi --module tasks.ca.statcan.data tasks.ca.statcan.data.DownloadData --resolution da_ --survey census
Starting bigmetadata_postgres_1 ... done
Starting bigmetadata_bigmetadata_daemon_1 ... done
INFO: Starting new HTTP connection (1): bigmetadata_daemon
INFO: Informed scheduler that task   tasks.ca.statcan.data.DownloadData_da__census_c30a093026   has status   PENDING
INFO: Done scheduling tasks
INFO: Running Worker with 1 processes
INFO: [pid 1] Worker Worker(salt=846184023, workers=1, host=606fcbf120ed, username=root, pid=1) running   tasks.ca.statcan.data.DownloadData(resolution=da_, survey=census)
2018-02-19 13:37:48,706 Downloading http://www12.statcan.gc.ca/census-recensement/2011/dp-pd/prof/details/download-telecharger/comprehensive/comp_download.cfm?CTLG=98-316-XWE2011001&FMT=CSV1501
INFO: Downloading http://www12.statcan.gc.ca/census-recensement/2011/dp-pd/prof/details/download-telecharger/comprehensive/comp_download.cfm?CTLG=98-316-XWE2011001&FMT=CSV1501
ERROR: [pid 1] Worker Worker(salt=846184023, workers=1, host=606fcbf120ed, username=root, pid=1) failed    tasks.ca.statcan.data.DownloadData(resolution=da_, survey=census)
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 194, in run
    new_deps = self._run_get_new_deps()
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 131, in _run_get_new_deps
    task_gen = self.task.run()
  File "/bigmetadata/tasks/base_tasks.py", line 423, in run
    self.uncompress()
  File "/bigmetadata/tasks/base_tasks.py", line 452, in uncompress
    z.extractall(path='{output}'.format(output=self.output().path))
  File "/usr/lib/python3.5/zipfile.py", line 1347, in extractall
    self.extract(zipinfo, path, pwd)
[...]

PR #432, opened 2018-02-20
Fix unzip fallback dir

juanignaciosl:
This is the complete fix for #427 after #431. I had only fixed the first step (sorry about that), so when I run `make ca-census-all` it failed at `SplitAndTransposeData` step. With the right output dir, it works.


antoniocarlon:
Tested with two different Canada Census import:
 - `make -- run ca.statcan.data.Census --resolution da_ --survey census --topic t010` (deflate64)
 - `make -- run ca.statcan.data.Census --resolution cma --survey census --topic t010`

Issue #428, opened 2018-02-19
Repository for DO ETL files


juanignaciosl:
Currently, ETL depends upon 3rd party files. If they change (see #427), running fails. We should create a file repository for the ETL so repeatability is guaranteed. It should, at least...
- store the last version that worked ("development branch").
- be able to store the tuple (extension version, ETL version, data): that would make the dump reproducible.

Let's discuss implementation alternatives, @ethervoid @antoniocarlon

ethervoid:
PR #440

PR #440, opened 2018-02-28
Repository for DO ETL files

antoniocarlon:
Creating a repository to store the DO ETL files

Closes https://github.com/CartoDB/bigmetadata/issues/428


antoniocarlon:
**Note:** The task `us.census.tiger.DownloadTigerGeography` is not managed with the repo yet because currently the ftp (i.e. `ftp://ftp2.census.gov/geo/tiger/TIGER2014`) is not working (timeout).

antoniocarlon:
**Notes for the acceptance:** Test the following tasks:

 - [ ] au-geo
 - [ ] au-all
 - [ ] br-geo
 - [ ] br-census
 - [ ] ca-geo
 - [ ] ca-nhs-all
 - [ ] ca-census-all
 - [ ] es-cnig
 - [ ] es-ine
 - [ ] eu-geo
 - [ ] eu-data
 - [ ] fr-geo
 - [ ] fr-insee
 - [ ] fr-income
 - [ ] mx-geo
 - [ ] mx-census
 - [ ] uk-geo
 - [ ] uk-census
 - [ ] us-tiger
 - [ ] us-bls
 - [ ] us-acs
 - [ ] us-lodes
 - [ ] us-spielman
 - [ ] us-enviroatlas
 - [ ] us-huc
 - [ ] us-zillow

juanignaciosl:
Oh, I thought that this had already been deployed. Is this blocked? Could this be pushed forward?

antoniocarlon:
It needs CR/Acceptance. It's not blocked and as soon as it's green we can/should move it to production

tasks/base_tasks.py
```diff
@@ -1592,3 +1610,133 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+    def run(self):
+        session = current_session()
+
+        create_schema = '''
+                        CREATE SCHEMA IF NOT EXISTS "{schema}"
+                        '''.format(schema=self.output().schema)
+        session.execute(create_schema)
+
+        create_table = '''
+                        CREATE TABLE IF NOT EXISTS "{schema}".{table} (
```
juanignaciosl: Minor: I'd quote consistently, either both schema and table or none.

tasks/base_tasks.py
```diff
@@ -1592,3 +1610,133 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+                            checksum TEXT,
+                            url TEXT,
+                            path TEXT,
+                            added TIMESTAMP,
+                            PRIMARY KEY(id, version)
+                        );
+                        '''.format(schema=self.output().schema,
+                                   table=self.output().tablename)
+        session.execute(create_table)
+        session.commit()
```
juanignaciosl: Probably caused by my ignorance: I see a `commit`, but not an explicit transaction beginning, or error handling with a proper `rollback`. How come?

tasks/au/data.py
```diff
@@ -46,15 +45,21 @@ class DownloadData(DownloadUnzipTask):
     state = Parameter()
     header = Parameter()
 
+    def version(self):
+        return 1
+
+    def requires(self):
+        return RepoFile(resource_id=self.task_id,
```
juanignaciosl: Minor: there are 31 creations of `RepoFile` that all have `resource_id` and `version` taken from the Luigi task. What about creating a `repofile_from_task` factory method? Feel free to call me _javero_ 😅

tasks/au/data.py
```diff
@@ -46,15 +45,21 @@ class DownloadData(DownloadUnzipTask):
...
     def download(self):
-        urllib.request.urlretrieve(url=URL.format(
-                               year=self.year,
-                               profile=self.profile,
-                               resolution=self.resolution,
-                               state=self.state,
-                               header=self.header
-                           ),
-                           filename=self.output().path + '.zip')
+        self.output().makedirs()
```
juanignaciosl: As `copyfile` receives the output file and recreating the needed directory structure if needed is a safe operation, I'd consider moving the equivalent to `self.output().makedirs()` into `copyfile`.

tasks/base_tasks.py
```diff
@@ -747,6 +750,21 @@ def mark_done(self):
...
 
 
+class LoadPostgresFromZipFile(TempTableTask):
+
+    def load_from_zipfile(self, zipfile):
+        shell('gunzip {zipfile} -c | grep -v default_tablespace | psql'.format(
+            zipfile=zipfile))
+        self.mark_done()
+
+    def mark_done(self):
```
juanignaciosl: Could you please explain me what's this for? I guess that it's Luigi-related, but...

tasks/base_tasks.py
```diff
@@ -1592,3 +1610,133 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+
+
+class RemoteDownloader:
+    def download(self, url, output_path):
+        raise NotImplementedError
+
+
+class BaseDownloader(RemoteDownloader):
+    def download(self, url, output_path):
+        urllib.request.urlretrieve(url, output_path)
```
juanignaciosl: Neither `BaseDownloader` nor `RemoteDownloader` rely on object state, and using them for inheritance leads to checks such as `if not issubclass(self.downloader.__class__, RemoteDownloader):`. What if those classes were removed and this `download` method became a normal function?

tasks/base_tasks.py
```diff
@@ -1592,3 +1610,133 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+        digest = self._retrieve_remote_file()
+        self._to_db(self.resource_id, self.version, digest, self.url, self.output().path)
+
+    def _create_filepath(self):
+        return self._build_path(str(uuid.uuid4()))
+
+    def _build_path(self, filename):
+        return os.path.join(self._repo_dir, self.resource_id, str(self.version), filename)
+
+    def _retrieve_remote_file(self):
```
juanignaciosl: Minor: making `_retrieve_remote_file` directly return the digest feels random, but, as it's private, feel free to ignore.

tasks/base_tasks.py
```diff
@@ -1592,3 +1610,133 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+
+    # https://stackoverflow.com/questions/3431825/generating-an-md5-checksum-of-a-file
+    def _digest(self, file):
+        hash_md5 = hashlib.md5()
+        with open(file, "rb") as f:
+            for chunk in iter(lambda: f.read(4096), b''):
+                hash_md5.update(chunk)
+        return hash_md5.hexdigest()
+
+    def _from_db(self, resource_id, version):
```
juanignaciosl: Have you considered moving these methods related to the table to `CreateRepoTable`? As they only depend on `resource_id` and `version`, `self.input().repo_entry(resource_id, version)` might make sense: it keeps table structure within that file.

tasks/base_tasks.py
antoniocarlon: Quoting is dangerous. The schema name needs to be quoted because the schemas almost always include a dot (`es.ine`), and thus in all the DO functions and ETL tasks we use quotes for it. But we never use quotes for table names and if we quote it here we would need to modify all the functions and tasks to add quotes for table names (a table created `create table myschema."MyTable"` will return an error `table doesn't exist` if accessed without quotes `select * from myschema.MyTable`)

juanignaciosl: Oh, I was just plain wrong about this, sorry! We've had a lot of problems with table naming in Builder and I recalled ending up quoting pretty much everything.

tasks/base_tasks.py
antoniocarlon: Luigi uses the target in the `output` function to know when the task can be marked as `DONE`. In this case, the schema is loaded from a URL (or a zip file after this PR) that contains all the SQL statements needed to create the schema, tables and insert the data instead of being just a plain CSV file, so we can't control the table/schema name and we create a fake table from the  output to know when all the data has been loaded.
It's far from being perfect but as it involved Tiger/ACS changed I have preferred not to change it ATM.

tasks/base_tasks.py
antoniocarlon: You are right. When I created the function I didn't want to introduce another indirection but taking a look after a while it feels a bit weird. I would glad to discuss a change if you think it's needed.

tasks/base_tasks.py
juanignaciosl: Ok, but how come the table must be _recreated_ on `mark_done`? If that's just a defensive measure you could use `IF NOT EXISTS` instead.

In addition, maybe adding the explanation as a comment might be nice, as this is not obvious for a reader.

tasks/base_tasks.py
juanignaciosl: Somehow following the command-query separation principle I'd remove `return self._digest(self.output().path)` from `_retrieve_remote_file` and change

```python
digest = self._retrieve_remote_file()
```

into

```python
self._retrieve_remote_file()
digest = self._digest(self.output().path)
```

Although it introduces sequence coupling (you must run `_digest` after `_retrieve_remote_file`), I think that it makes sense from an OO and Python point of view. And, in addition, that coupling is there anyway, as `self.output().path` is passed to `_to_db`.

tasks/base_tasks.py
antoniocarlon: You are right... part of my heart still runs on Java. Changing it :)

tasks/base_tasks.py
antoniocarlon: The Luigi structure is a bit weird: `self.input()` doesn't return `CreateRepoTable` but its output (`PostgresTarget`). We could create a more complex structure to move these functions to `CreateRepoTable` but I think it's not worth it...

tasks/au/data.py
antoniocarlon: `copyfile` is a function from `shutil` (included in _The Python Standard Library_)

juanignaciosl: Well, I'd wrap it in a `copyfile_makedirs` function 😅 (sorry about not realizing that it's an imported function, I still don't get used to that :_) ).

tasks/au/data.py
antoniocarlon: I considered doing something like that but I haven't done it because the current structure is our first approach and I would expect us to refine it a bit. I would wait until we have the structure completely clear, but we can discuss it if you think that it won't change (or at least won't change _a lot_)

juanignaciosl: You're closer to the code, so up to you. I think that a factory function with named parameters and defaults should resits changes nicely, but whatever you decide will be fine.

tasks/base_tasks.py
antoniocarlon: The transaction is managed by SQLAlchemy (`session = current_session()`). But you are right, a `except: session.rollback()` won't hurt

tasks/base_tasks.py
juanignaciosl: Oh, I didn't remember that 👍

tasks/base_tasks.py
antoniocarlon: It's a small table with one row that stores the creation time of the table. If we use `IF NOT EXISTS` we would need to do an `UPDATE` or `INSERT` afterwards. This `LoadPostgresFromZipFile` just replicates the existing `LoadPostgresFromURL` but using an existing file instead of downloading it.
I'm adding a comment to make it clear.

tasks/base_tasks.py
antoniocarlon: Totally agree. I couldn't see the forest for the trees. Changed it

antoniocarlon:
@juanignaciosl another round?

tasks/base_tasks.py
```diff
@@ -1592,3 +1592,127 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+
+def base_downloader(url, output_path):
+    urllib.request.urlretrieve(url, output_path)
+
+
+class RepoFile(Task):
+    resource_id = Parameter()
+    url = Parameter()
+    version = IntParameter()
+    downloader = Parameter(default=base_downloader)
```
ethervoid: Is this really needed? Parameters in luigi are not like arguments in the constructor, they define the instance name of the class so if this is needed maybe we need to add the [significant](http://luigi.readthedocs.io/en/stable/parameters.html#insignificant-parameters) property

tasks/base_tasks.py
```diff
@@ -1592,3 +1592,127 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+        hash_md5 = hashlib.md5()
+        with open(file, "rb") as f:
+            for chunk in iter(lambda: f.read(4096), b''):
+                hash_md5.update(chunk)
+        return hash_md5.hexdigest()
+
+    def _from_db(self, resource_id, version):
+        checksum, url, path = None, None, None
+        query = '''
+                SELECT checksum, url, path FROM "{schema}".{table}
```
ethervoid: We should start moving this stuff to be used by sqlalchemy. This is a regular query to a table. Not now, because we dont have define the model for that table but we should avoid plain sql

tasks/base_tasks.py
```diff
@@ -1592,3 +1592,127 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+                           resource_id=resource_id,
+                           version=version)
+        result = current_session().execute(query).fetchone()
+        if result:
+            checksum, url, path = result
+
+        return checksum, url, path
+
+    def _to_db(self, resource_id, version, checksum, url, path):
+        query = '''
```
ethervoid: Same here

tasks/base_tasks.py
```diff
@@ -1592,3 +1592,127 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+        path = self._from_db(self.resource_id, self.version)[2]
+        if not path:
+            if not self._path:
+                self._path = self._create_filepath()
+            path = self._path
+
+        return path
+
+    def complete(self):
+        deps = self.deps()
```
ethervoid: Hate this too but unless we find a way to get rows from a table we are stick to this kind of `complete` methods which are error-prone

tasks/base_tasks.py
```diff
@@ -1592,3 +1592,127 @@ def collect_meta_wrappers(test_module=None, test_all=True):
...
+        return self._build_path(str(uuid.uuid4()))
+
+    def _build_path(self, filename):
+        return os.path.join(self._repo_dir, self.resource_id, str(self.version), filename)
+
+    def _retrieve_remote_file(self):
+        self.downloader(self.url, self.output().path)
+
+    # https://stackoverflow.com/questions/3431825/generating-an-md5-checksum-of-a-file
+    def _digest(self, file):
```
ethervoid: I'll move this method to a utility class lib/util.py or lib/file_utils.py or something

tasks/util.py
```diff
@@ -421,3 +422,16 @@ def grouper(iterable, n, fillvalue=None):
...
+
+
+def copyfile(src, dst):
+    # From http://luigi.readthedocs.io/en/stable/_modules/luigi/local_target.html#LocalTarget.makedirs
+    normpath = os.path.normpath(dst)
+    parentfolder = os.path.dirname(normpath)
+    if parentfolder:
+        try:
+            os.makedirs(parentfolder)
+        except OSError:
```
ethervoid: Maybe add here log could be useful

tasks/uk/cdrc.py
```diff
@@ -36,18 +36,30 @@ def tags(self):
                    ]
 
 
-class DownloadOutputAreas(DownloadUnzipTask):
+def cdrc_downloader(url, output_path):
+    if 'CDRC_COOKIE' not in os.environ:
+        raise ValueError('This task requires a CDRC cookie. Put it in the `.env` file\n'
+                         'e.g: CDRC_COOKIE=\'auth_tkt="00000000000000000username!userid_type:unicode"\'')
+    shell('wget --header=\'Cookie: {cookie}\' -O {output} {url}'.format(
```
ethervoid: Hate this

tasks/eu/geo.py
```diff
@@ -31,28 +31,33 @@ def tags(self):
         ]
 
 
-class DownloadGeographies(Task):
+def geographies_downloader(url, output_path):
+    referer = 'http://www.eurogeographics.org/content/euroglobalmap-opendata?sid=10868'
+    shell("wget -O {output} --referer='{referer}' '{url}'".format(
```
ethervoid: Hate this

tasks/base_tasks.py
```diff
@@ -727,18 +730,15 @@ def after_copy(self):
         pass
 
 
-class LoadPostgresFromURL(TempTableTask):
+class LoadPostgresFromZipFile(TempTableTask):
```
ethervoid: I wouldn't delete LoadPosgresFromURL yet

tasks/base_tasks.py
```diff
@@ -727,18 +730,15 @@ def after_copy(self):
...
-    def load_from_url(self, url):
-        '''
-        Load psql at a URL into the database.
-
-        Ignores tablespaces assigned in the SQL.
-        '''
-        shell('curl {url} | gunzip -c | grep -v default_tablespace | psql'.format(
-            url=url))
+    def load_from_zipfile(self, zipfile):
+        shell('gunzip {zipfile} -c | grep -v default_tablespace | psql'.format(
```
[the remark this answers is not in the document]
antoniocarlon: I know and I thought about when I was developing it, but as this change is big enough I didn't want to add a new variable to the acceptance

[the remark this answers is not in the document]
ethervoid: Yep this one is a tricky one for us, so we should be aware that we have to remove it but not now

antoniocarlon:
@ethervoid round 2?

juanignaciosl:
`br-census` fails because of encoding. Example:

```
2018-05-22 11:21:29,606 [INFO]: [pid 20] Worker Worker(salt=834209542, workers=1, host=dfb6d9961b04, username=root, pid=20) running   tasks.br.data.ImportData(state=ba, tablename=Basico)
2018-05-22 11:21:43,434 [ERROR]: [pid 20] Worker Worker(salt=834209542, workers=1, host=dfb6d9961b04, username=root, pid=20) failed    tasks.br.data.ImportData(state=ba, tablename=Basico)
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 194, in run
    new_deps = self._run_get_new_deps()
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 131, in _run_get_new_deps
    task_gen = self.task.run()
  File "/bigmetadata/tasks/base_tasks.py", line 705, in run
    coldef=', '.join(['"{}" {}'.format(c[0], c[1]) for c in self.coldef()])
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/orm/session.py", line 1044, in execute
    bind, close_with_result=True).execute(clause, params or {})
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/base.py", line 947, in execute
    return meth(self, multiparams, params)
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/sql/elements.py", line 262, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/base.py", line 1055, in _execute_clauseelement
    compiled_sql, distilled_params
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/base.py", line 1191, in _execute_context
    context)
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/base.py", line 1389, in _handle_dbapi_exception
    util.reraise(*exc_info)
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/util/compat.py", line 186, in reraise
    raise value
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/base.py", line 1184, in _execute_context
    context)
  File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/default.py", line 462, in do_execute
    cursor.execute(statement, parameters)
UnicodeEncodeError: 'ascii' codec can't encode character '\xf5' in position 91: ordinal not in range(128)
```

You can reproduce it easily with `make -- docker-run br.data.ImportData --state ba --tablename Basico`. 

Database is UTF8, I don't know how come SQL Alchemy is assuming ascii:

```
gis=# show server_encoding;
 server_encoding
-----------------
 UTF8
(1 row)
```

juanignaciosl:
Ok, database encoding was actually wrong, because the check that I sent previously was run on `postgres10`. `postgres` (9.5) encoding was wrong. @antoniocarlon recreated the postgres directory in the docker and encoding was right again 🤷‍♂️

juanignaciosl:
:rock

ethervoid:
Please fix the conflicts and merge this :)

Issue #429, opened 2018-02-19
Create a task to rebuild all the coutries every X days


ethervoid:
Create a task with the following objectives:

- [x] Clean all the data from the countries passed or hardcoded in a list: downloaded files in tmp, schemas for intermediate table, data in observatory and metadata tables
- [x] Execute, using a docker machine for every task, all the defined tasks in one list. For example es-all, ca-all, eu-all and so on
- [x] Generate logs for every task
- [ ] [Nice to have] Analyze the logs, once the task has finished, and compile the errors. Is a nice to have because at first could be done by hand
- [x] Send email/notification when it finish --> Need a webhook for slack

PR #441, opened 2018-02-28
Rebuild all script

ethervoid:
Related #429


README.md
```diff
@@ -75,6 +76,27 @@ For example, to run QCEW numbers for one quarter:
...
+usage: python3 watch_containers.py [-h] [--since SINCE]
+                                   [--pooling-time POOLING_TIME]
+                                   [--notification-channel {stdout,slack,logfile}]
+                                   name
+```
+
+Let me explain the options:
+
+* **name**: Name, complete or a part of it, of the container to watch
+* **since**: [optional] Date in the `YYYY-mm-dd HH:MM:SS` format since the script starts looking for docker containers to watch. Default value is the now()
```
antoniocarlon: The `StartedAt` property of the container uses UTC time. We should default it to `utcnow()` (or something like `datetime.utcnow()-timedelta(hours=1)`)

scripts/watch_containers.py
```diff
@@ -0,0 +1,63 @@
...
+        active_containers = get_active_containers(client, api_client, name, since)
+        if not active_containers:
+            print("No more active containers. Exiting...")
+            break
+        else:
+            print("Still running containers. Sleeping...")
+            time.sleep(pooling_time)
+
+    for container in containers:
+        exit_code = api_client.inspect_container(container.id)['State']['ExitCode']
```
antoniocarlon: If the container is stopped and removed (for example doing `docker-compose down`) this will fail. We could add a new list `killed_containers` and add these containers there to show them on the stats too

ethervoid:
``` ❯ python3 scripts/watch_containers.py bigmetadata_bigmetadata_run_ --since '2018-03-02 16:28:00'       [17:30:48]
Still running containers. Sleeping...
No more active containers. Exiting...
Finished. Total 1 -- Successful 0 -- Failed 0 -- Deleted 1. It tooks 3 minutes
>>> elapsed time 1m1s   ```

ethervoid:
@antoniocarlon Another round?

PR #445, opened 2018-03-05
Added target for docker-run

antoniocarlon:
- Added make target for `docker-run`
 - Updated documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ endif
 run:
 	python3 -m luigi $(SCHEDULER) --module tasks.$(MOD_NAME) tasks.$(TASK)
 
+docker-run:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,6 +26,9 @@ endif
 run:
 	python3 -m luigi $(SCHEDULER) --module tasks.$(MOD_NAME) tasks.$(TASK)
 
+docker-run:
+	docker-compose run -d -e LOGGING_FILE=etl_$(MOD_NAME).log bigmetadata luigi --module tasks.$(MOD_NAME) tasks.$(TASK)
```
ethervoid: What if we add for the LOGGING_FILE `etl_$(MOD_NAME)_$(TASK)`?

antoniocarlon: The task also includes the parameters in the form of `--parameter value` that's why I'm using the module name

ethervoid: Ahh is true! Cool!

PR #446, opened 2018-03-05
Multiple additions

ethervoid:
- Logger for the watcher
- Slack notifications integration

### DIFF
--- a/scripts/watch_containers.py
+++ b/scripts/watch_containers.py
@@ -1,19 +1,25 @@
 import docker
 import time
 import argparse
+import os
+import requests
+import logging
+import logging.config
+import json
 from datetime import datetime
 from datetime import timedelta
 
 STDOUT = "stdout"
 SLACK = "slack"
-LOGFILE = "logfile"
-NOTIFICATION_CHANNELS = [STDOUT, SLACK, LOGFILE]
+NOTIFICATION_CHANNELS = [STDOUT, SLACK]
 DOCKER_SOCKET_URL = 'unix://var/run/docker.sock'
 DEFAULT_POLLING_TIME = 60
 DOCKER_STATUS_RUNNING = 'running'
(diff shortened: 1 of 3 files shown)


scripts/watch_containers.py
```diff
@@ -38,9 +44,9 @@ def watch_containers(name, since=datetime.utcnow(), polling_time=DEFAULT_POLLING
                 successful_containers.append(container)
         except docker.errors.APIError:
             deleted_containers.append(container)
-    finish_time = datetime.utcnow() - since
+    finish_time = datetime.utcnow() - init_time
     finish_time_in_minutes = finish_time / timedelta(minutes=1)
-    message = "Finished. Total {} -- Successful {} -- Failed {} -- Deleted {}. It tooks {} minutes".format(len(containers), len(successful_containers), len(failed_containers), len(deleted_containers), int(finish_time_in_minutes))
+    message = "Finished build. Total {} -- Successful {} -- Failed {} -- Deleted {}.\nIt tooks {} minutes".format(len(containers), len(successful_containers), len(failed_containers), len(deleted_containers), int(finish_time_in_minutes))
```
antoniocarlon: Minor typo: should be `It took {} minutes`

scripts/watch_containers.py
```diff
@@ -56,20 +62,30 @@ def get_active_containers(client, api_client, name, since):
 
 def send_notification(channel, message):
     if channel == STDOUT:
-        print(message)
+        logger.info(message)
     elif channel == SLACK:
-        raise NotImplementedError('Slack notification is not implemented yet')
-    elif channel == LOGFILE:
-        raise NotImplementedError('Log file notification is not implemented yet')
+        if 'SLACK_WEBHOOK' not in os.environ:
```
antoniocarlon: Would it make sense to fall back to `STDOUT` if there is no `SLACK_WEBHOOK`?

scripts/watch_containers.py
ethervoid: I was doubting about it hhehe. Thx

Issue #336, opened 2017-10-03
Redo/Refactor/Fix metadata logic


ethervoid:
Related #294 

During the related task I've found some problems with the metadata tasks:

* Some tasks like `ColumnTask` is not idempotent and does operations that are not recommended by Luigi like:
  - Have logic in the `complete` method
  - Call and get the `output` objects in the `run` method
  - Call `complete` from `output` method
  - They're not idempotent because they don't have a clear output
* Also I've found that the ORM model is more complex than needed so we probably have to check if we can simplify it
* We have more tasks without an idempotent `output` so we have to fix that too

ethervoid:
#### Meeting summary and conclusions

Yesterday we had a meeting to discuss what we should do with [this ticket](https://github.com/CartoDB/bigmetadata/issues/294) and if we block it or continue working on it.

The conclusion was:
  - Block the ticket and stop spending time on it because we have to fix first some other issues like the metadata one (that is covered in this ticket)
  - Create and be aware of this ticket but don't start working on it immediately because the problem is not a blocking one.

juanignaciosl:
As a specific conclusion of "be aware of this ticket", we agreed on considering creating a simpler task class hierarchy when we tackle the "declarative ETL" task in the future.

PR #447, opened 2018-03-05
Revert parallel execution for Luigi

ethervoid:
After some working with it, we're getting odd behaviour using the
parallel execution so is better to avoid it by now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,7 @@ docs:
 	docker-compose run --rm bigmetadata /bin/bash -c 'cd docs && make html'
 
 tiles:
-	make run-parallel util.GenerateAllRasterTiles
+	make run util.GenerateAllRasterTiles
 
 meta:
 	make run -- carto.OBSMetaToLocal --force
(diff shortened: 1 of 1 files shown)


ethervoid:
What do you think @antoniocarlon ?

juanignaciosl:
I must admit that I don't have enough information. What "odd behaviour" are we getting and what do we lose if we remove it?

antoniocarlon:
In our experience parallel execution + our implementation of the SQLAlchemy models don't behave well together.
If you have found more issues with that I would avoid using it until we have time to take a more deep look into it.

ethervoid:
@juanignaciosl, for example, Eurostat works in single worker execution but if you put it in parallel execution it starts messing with the DB sessions and fails.

We have plenty of cases about problems with SQL Sessions and until we don't cleanse/revamp the models part using parallel execution is not going to be an option

juanignaciosl:
@ethervoid thanks! I understand that you're well aware of the problems, but readers of this PR (that doesn't have any linked issue) should also be informed in order to understand it 👍

ethervoid:
More info #336

PR #482, opened 2018-04-17
Remove unused parameters

ethervoid:
This improve the name of the downloaded datasets so we could know
for which measure and type is for


antoniocarlon:
Tested doing `make -- docker-run us.zillow.Zillow --geography Zip --year 2016 --month 2` and checked that everything is working as expected --> merging!

PR #495, opened 2018-05-21
Xyz tables generation

ethervoid:


tasks/tiler/do.py
```diff
@@ -0,0 +1,110 @@
...
+                    area_ratio NUMERIC,
+                    {},
+                    CONSTRAINT xyzusall_pk PRIMARY KEY (x,y,z,geoid)
+                )'''.format(table_schema['table_name'], ", ".join(cols_schema))
+            session.execute(sql_table)
+            session.commit()
+
+    def _get_config_data(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        with (open('{}/us_all.json'.format(dir_path))) as f:
```
ethervoid: We can pass it as argument for example

tasks/tiler/do.py
```diff
@@ -0,0 +1,122 @@
...
+            for y in range(0, (pow(2, zoom) + 1)):
+                geography = self._get_geography_level(zoom)
+                sql_tile = '''
+                    INSERT INTO {table}
+                    (SELECT {x}, {y}, {z}, mvtgeom, {recordset}
+                    FROM cdb_observatory.OBS_GetMCDOMVT({z},{x},{y},'{geography}',ARRAY['{docols}']::TEXT[],ARRAY['{mccols}']::TEXT[]));'''.format(table=table_name, x=x, y=y, z=zoom, geography=geography, recordset=", ".join(recordset), docols="', '".join(do_columns), mccols="', '".join(mc_columns))
+                session.execute(sql_tile)
+            session.commit()
+
+    def _get_geography_level(self, zoom):
```
antoniocarlon: As the task receives the `zoom` and the `geography` as parameters, it doesn't make much sense getting the geography level from the zoom here. It would be better if we add a dictionary to get the geography id (for example `us.census.tiger.state`) from the geography (`state`)

ethervoid: Yep, I left the duplication to be tackle after but change it please :)

PR #502, opened 2018-05-29
Multi-country but for US only by now

ethervoid:


tasks/tiler/us.py
```diff
@@ -0,0 +1,50 @@
...
+        return GEOGRAPHY_LEVELS[level]
+
+
+class AllUSXYZTables(WrapperTask):
+
+    def requires(self):
+        for zoom in range(0, 5):
+            yield XYZTables(zoom_level=zoom, geography=self._get_geography_level(zoom))
+
+    def _get_geography_level(self, zoom):
```
juanignaciosl: I'd replace these conditionals with a `geography_level_by_zoom` dict or array. What do you think?

tasks/tiler/xyz.py
```diff
@@ -14,35 +15,24 @@
...
-# rects are assumed to be [xmin, ymin, xmax, ymax]
-def bboxes_intersect(rect1, rect2):
-    return not (rect2[0] > rect1[2]
-                or rect2[2] < rect1[0]
-                or rect2[3] < rect1[1]
-                or rect2[1] > rect1[3])
-
-
-class XYZUSTables(Task):
+class TilerXYZTableTask(Task):
```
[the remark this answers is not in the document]
juanignaciosl: ☝️

tasks/tiler/xyz.py
```diff
@@ -187,22 +177,3 @@ def output(self):
             table_schema = self._get_table_schema(table_config)
             targets.append(PostgresTarget(table_schema['schema'], table_schema['table_name'], where='z = {}'.format(self.zoom_level)))
         return targets
-
-
-class AllXYZTables(WrapperTask):
```
juanignaciosl: This is mentioned in the Makefile, it should change.

tasks/tiler/us.py
ethervoid: that is wors for flexibility. At first, I did that but I ended having a custom dictionary for every subclass which is worse to implement in the future

tasks/tiler/xyz.py
ethervoid: do.py is removed, now is called xyz

ethervoid: changed in the makefile

tasks/tiler/us.py
ethervoid: Oh I didn't see is the other function sorry. Yes, I tried to do because is a range of integers you end having a function to gather the data or having an array with 14 elements. SUggestions?

juanignaciosl: I'd say that an array is a good fit, because it has implicit semantics about key type and failure if zoom is not supported. It might look ugly but it's precise, clear, small and provides O(1) reads. I based my suggestion on a "Writing idiomatic Python" book recommendation, but as there are zoom ranges, the improvement is not _that_ clear. Anyway, it's really minor, if you prefer the `elif` approach, it's fine 👍

PR #513, opened 2018-06-11
Add State and County to zillow

ethervoid:


ethervoid:
Blocked because we have to realize/add how to add the geoid to the zillow data to create the relationship between geometry and data using geoid (zillow only has the name of the region).

Ie. for county we have allen but we need the geoid for that county to create the relationship with tiger geometries

PR #514, opened 2018-06-11
Tiger blocks simplification by state

### LINKED ISSUE #512: Tiger blocks simplification by state
The Tiger blocks layer is huge and we can't simplify it using mapshaper because of OOMs. We are simplifying it using the PostGIS built-in functions (`ST_SimplifyVW`) but we are losing the topology that mapshaper ensures:

simplification-do

To avoid the problem and achieve a layer with some topology we need to divide and simplify the layer by state using mapshaper.

antoniocarlon:
Simplifying the Tiger blocks by state to take advantage of the mapshaper topologically correct simplification
Closes #512


tasks/us/census/tiger.py
```diff
@@ -385,6 +388,95 @@ def output(self):
...
+        return simplifications
+
+    def run(self):
+        session = current_session()
+        for _, table in self.input().items():
+            query = '''
+                    CREATE TABLE IF NOT EXISTS {output} AS
+                    SELECT *
+                    FROM "{schema_input}".{table_input}
+                    WHERE 1=0
```
ethervoid: Are you only creating the table without rows?

tasks/us/census/tiger.py
```diff
@@ -385,6 +388,95 @@ def output(self):
...
+                    '''.format(schema_input=table.schema,
+                               table_input=table.tablename,
+                               output=self.output().table)
+            session.execute(query)
+
+            query = '''
+                    INSERT INTO {output}
+                    SELECT *
+                    FROM "{schema_input}".{table_input}
```
ethervoid: Why do you make it in two steps?

antoniocarlon: These `create` + `insert` are inside a `for` loop that iterates over all the simplified tables. This inserts into the output table every iteration but creates the table only on the first one (`IF NOT EXISTS`)

ethervoid: offline -> `input_table` has all the same columns. First create table creates a table with the output table not with the input table and then start with input

PR #522, opened 2018-06-29
Adding Category and Date (month) columns to XYZ generation

### LINKED ISSUE #519: Adapt the XYZ to the requirements (categories/dates)
Adapt the XYZ to store the required fields (categories, dates, ...)

antoniocarlon:
n-plicates the columns based on the categories and dates (month) of the data

Closes #519

### DIFF
--- a/tasks/tiler/xyz.py
+++ b/tasks/tiler/xyz.py
@@ -40,7 +40,7 @@ class TilerXYZTableTask(Task):
             table_schema = self._get_table_schema(table_config)
             table_bboxes = table_config['bboxes']
             self._create_schema_and_table(table_schema)
-            self._generate_and_insert_tiles(self.zoom_level, table_schema, table_config['columns'], table_bboxes)
+            self._generate_and_insert_tiles(self.zoom_level, table_schema, table_config, table_bboxes)
 
     def _create_schema_and_table(self, table_schema):
         session = current_session()
@@ -70,11 +70,17 @@ class TilerXYZTableTask(Task):
 
     def _get_table_schema(self, table_config):
         columns = {}
-        for dataset, col_data in table_config['columns'].items():
-            columns[dataset] = []
-            for column in col_data:
-                nullable = '' if column['nullable'] else 'NOT NULL'
-                columns[dataset].append("{} {} {}".format(column['column_name'], column['type'], nullable))
+
+        columns['do'] = []
+        for do_column in table_config['columns']['do']:
+            nullable = '' if do_column['nullable'] else 'NOT NULL'
+            columns['do'].append("{} {} {}".format(do_column['column_name'], do_column['type'], nullable))
+
+        columns['mc'] = []
+        for mc_category in table_config['mc_categories']:
+            for mc_column in table_config['columns']['mc']:
+                nullable = '' if mc_column['nullable'] else 'NOT NULL'
+                columns['mc'].append("{} {} {}".format(mc_column['column_name'] + '_' + mc_category['id'], mc_column['type'], nullable))
         return {"schema": table_config['schema'], "table_name": table_config['table'], "columns": columns}
 
     def _tile_in_bboxes(self, zoom, x, y, bboxes):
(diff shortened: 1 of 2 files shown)


tasks/tiler/xyz.py
```diff
@@ -34,24 +35,29 @@ def get_config_file(self):
...
+
+    def _get_table_name(self, table_schema):
+        return table_schema['table_name'] + '_' + self.month
+
+    def _get_mc_dates(self):
+        session = current_session()
+        query = '''
+                SELECT cdb_observatory.OBS_GetMCDates('{schema}', '{geo_level}', '{month}');
+                '''.format(schema='us.mastercard', geo_level=self.geography, month=self.month)
+        return session.execute(query).fetchone()[0]
```
ethervoid: Here we're returning all the dates. Why don't we only use the month? O you want to leave this open in order to ask for a defined date for example 10/10/2017?

tasks/tiler/xyz.py
```diff
@@ -34,24 +35,29 @@ def get_config_file(self):
...
 
     def _create_schema_and_table(self, table_schema):
         session = current_session()
         session.execute('CREATE SCHEMA IF NOT EXISTS tiler')
         cols_schema = []
-        table_name = "{}.{}".format(table_schema['schema'], table_schema['table_name'])
+        table_name = "{}.{}".format(table_schema['schema'], self._get_table_name(table_schema))
```
ethervoid: We should change this to `"\"{}\".\"{}\""` in order to use double quote with the postgis schema or table

tasks/tiler/xyz.py
```diff
@@ -61,6 +67,7 @@ def _create_schema_and_table(self, table_schema):
                 mvt_geometry Geometry NOT NULL,
                 geoid VARCHAR NOT NULL,
                 area_ratio NUMERIC,
+                area NUMERIC,
```
ethervoid: Why do we need the area now?

tasks/tiler/xyz.py
```diff
@@ -73,12 +80,23 @@ def _get_config_data(self):
...
-                nullable = '' if column['nullable'] else 'NOT NULL'
-                columns[dataset].append("{} {} {}".format(column['column_name'], column['type'], nullable))
+
+        columns['do'] = []
+        for do_column in table_config['columns']['do']:
+            nullable = '' if do_column['nullable'] else 'NOT NULL'
+            columns['do'].append("{} {} {}".format(do_column['column_name'], do_column['type'], nullable))
+
+        columns['mc'] = []
+        for mc_date in mc_dates:
```
ethervoid: Should not be this month instead of date? We are creating a column by date instead of by month

tasks/tiler/xyz.py
antoniocarlon: The MBTiles have the area and I wanted to return the exact same fields that they already have

tasks/tiler/xyz.py
antoniocarlon: Totally. On it

tasks/tiler/xyz.py
antoniocarlon: I would like to be able to ask for a given date (as it's not very well defined right now I wanted it to be a bit more open, but we can revisit this later if needed)

tasks/tiler/xyz.py
antoniocarlon: Yes, they need a column by each date. The input is the month and for that month we look for every date with that month and create a column with each date among those

ethervoid: So how many columns?

ethervoid:
Fix tests :D

tasks/tiler/xyz.py
antoniocarlon: Depends on the dates that we have in our mc data for a given month. Currently, 5 columns for each month, totalling `number_of_mccolumns x n_categories(currently 5) x n_dates(currently 5)`

ethervoid: Ok, this could lead to a ton of columns xD

antoniocarlon: yes, that's what we talked about before... splitting by months, splitting in several tables, adding all the dates into an array. If I recall correctly we decided to do it this way for this iteration and then think about the improvements if needed, isn't it?

ethervoid: Yep, it's just I thought it was by month instead of by date hehe

PR #529, opened 2018-07-11
Canada CSV generation

ethervoid:


tasks/au/data.py
```diff
@@ -394,11 +404,42 @@ def columns(self):
...
+            self.populate_general()
+
+    def populate_mb(self):
+        session = current_session()
+        column_targets = self.columns()
+        out_colnames = [oc.lower() for oc in list(column_targets.keys())]
+        in_colnames = ['mb.geom_id as region_id']
+        for ic in list(column_targets.keys()):
+            if ic != 'region_id':
+                in_colnames.append('round(cast(float8 ({ic} * (ST_Area(mb.the_geom)/ST_Area(sa1geo.the_geom))) as numeric), 2) as {ic}'.format(ic=ic.lower()))
```
antoniocarlon: Are MBs fully enclosed in SA1s?

tasks/geographica/ca/csv.py
```diff
@@ -0,0 +1,107 @@
...
+
+# TODO Add block level when we have ACS for block
+GEOGRAPHY_LEVELS = {
+    'GEO_PR': 'ca.statcan.geo.pr_',
+    'GEO_CD': 'ca.statcan.geo.cd_',
+    'GEO_DA': 'ca.statcan.geo.da_'
+}
+
+
+class Measurements2CSV(Task):
```
antoniocarlon: This task and the one for AU should be very similar. Could we refactor this so all the `Measurements2CSV` tasks inherit from a base task?

ethervoid: Yes, and the US too. I want to finish UK and then refactor them

tasks/au/data.py
ethervoid: Yep

PR #530, opened 2018-07-12
Import MC data rebuilding the CSV

antoniocarlon:
A new import mechanism for the MC CSV file based on generating a new CSV with the desired structure + COPY

https://github.com/CartoDB/observatory-extension/issues/342


antoniocarlon:
The performance is awful with this approach

Issue #503, opened 2018-05-29
Add more boundaries for the current countries


ethervoid:
Based on the recent work we're doing to server MVT tiles we found that we need more boundaries for some countries in order to have better aggregation and granularity for different zoom levels.

The ideal situation is to get boundaries that match for Tiger levels State, county, census_tract, block_group and block

* **UK**
  In UK we only have the CDRC boundaries but we should add more like for example country level

* **Spain**
  We only have the INE census parts but we could add State (country/admin0), Autonomous regions (county), Provinces and census

More to check in the future

juanignaciosl:
@ethervoid @antoniocarlon is there any current request or known need? As this issue is quite broad, it'd be great if we could make a breakdown of needed tasks.

antoniocarlon:
AFAIK, ATM only this is needed https://github.com/CartoDB/bigmetadata/issues/568
Closing this!

PR #535, opened 2018-07-20
Add Postcode Areas to UK

antoniocarlon:
Added Postcode Areas to UK interpolated from existing Census
#503


tasks/base_tasks.py
```diff
@@ -584,6 +584,7 @@ class Shp2TempTableTask(TempTableTask):
     '''
 
     encoding = Parameter(default='latin1', significant=False)
+    other_options = ''
```
juanignaciosl: `other_options` couples the usage to the precise `ogr2org` options. Although it might be seen as convenient, in the future it can become a problem not easy to get rid of. As it's only used to override source SRS, what do you think about adding an optional `override_srs` parameter instead, @ethervoid ?

juanignaciosl:
Acceptance notes:

Ok, after importing, `obs_getavailableboundaries` list it:

```
cartodb_dev_user_76f71c14-e13a-4893-b286-e4332eda6d27_db=> select boundary_id, tablename from OBS_GetAvailableBoundaries(cdb_latlng(51.51322957, -0.09115219));
           boundary_id            |                  tablename                   
----------------------------------+----------------------------------------------
 uk.cdrc.the_geom                 | obs_22ef53ead17e44478fb65e19006722b304dd7359
 uk.census.postcodeareas.the_geom | obs_46e2e0e2077d48594a5a1f73dead38d3d7019396
(2 rows)
```

... But then, I can't fetch them with `obs_getboundariesbygeometry`:

```
cartodb_dev_user_76f71c14-e13a-4893-b286-e4332eda6d27_db=> SELECT ROW_NUMBER() OVER () AS cartodb_id,
       the_geom,
       geom_refs,
       ST_TRANSFORM(the_geom, 3857) the_geom_webmercator
FROM OBS_GetBoundariesByGeometry(
  ST_GeomFromText('POLYGON((-8.4 49.8,-8.4 60.9,2.1 60.9,2.1 49.8,-8.4 49.8))', 4326),
  'uk.census.postcodeareas.the_geom')
;
 cartodb_id | the_geom | geom_refs | the_geom_webmercator 
------------+----------+-----------+----------------------
(0 rows)
```

I've checked that the geometries table is there, that geometries are well placed, and that the previous polygon overlaps the UK. 

Getting into `obs_getboundariesbygeometry`, I see that is calling `obs_getmeta` and that it returns `null`:

```
gis=# SELECT cdb_observatory.OBS_GetMeta(ST_GeomFromText('POLYGON((-8.4 49.8,-8.4 60.9,2.1 60.9,2.1 49.8,-8.4 49.8))', 4326), JSON_Build_Array(JSON_Build_Object('geom_id', 'uk.census.postcodeareas.the_geom', 'geom_timespan', null)));
 obs_getmeta 
-------------
 
(1 row)

```

juanignaciosl:
cc @ethervoid @antoniocarlon

juanignaciosl:
Running it with `PGSERVICE=postgres10 make docker-run carto.OBSMetaToLocal`, thanks! :+1:

juanignaciosl:
_While that's running..._

I have two general questions, unrelated to this PR:

What's the motivation of splitting data loading with metadata generation? Would it make sense creating the metadata on data loading?

The boundaries are actually there, I'm not sure how come (more) metadata is needed. Wouldn't that mean that, at least for `obs_getboundariesbygeometry`, `get_meta` performs "too many" checks?

(I haven't researched a lot, I want to know your thoughts about this).

ethervoid:
This is how it was originally created. Two steps, first data ingestion, and second metadata generation an exposure of the data. It's very probable that create it in two steps is easier for the tasks because we don't have to deal with the metadata in every task.

Also, we can use it to add data but not expose it until we generate the metadata so we use it in that way sometimes.

antoniocarlon:
Successfully executed in `edo-production`, waiting for the next dump

PR #552, opened 2018-08-16
Add ACS and TIGER 2016 data

antoniocarlon:
Added ACS and TIGER 2016 data
Closes https://github.com/CartoDB/bigmetadata/issues/392

### DIFF
--- a/tasks/us/census/acs.py
+++ b/tasks/us/census/acs.py
@@ -33,7 +33,7 @@ SAMPLE_5YR = '5yr'
 GEOGRAPHIES = [STATE, COUNTY, CENSUS_TRACT, BLOCK_GROUP, BLOCK, PUMA, ZCTA5, CONGRESSIONAL_DISTRICT,
                SCHOOL_DISTRICT_ELEMENTARY, SCHOOL_DISTRICT_SECONDARY, SCHOOL_DISTRICT_UNIFIED,
                CBSA, PLACE]
-YEARS = ['2015']
+YEARS = ['2015', '2016']
 SAMPLES = [SAMPLE_5YR]
 
 
@@ -82,9 +82,9 @@ class Quantiles(TableTask):
             'table': Extract(year=self.year,
                              sample=self.sample,
                              geography=self.geography),
-            'tiger': GeoidColumns(year='2015'),
-            'sumlevel': SumLevel(geography=self.geography, year='2015'),
-            'shorelineclip': ShorelineClip(geography=self.geography, year='2015')
+            'tiger': GeoidColumns(year=self.year),
+            'sumlevel': SumLevel(geography=self.geography, year=self.year),
+            'shorelineclip': ShorelineClip(geography=self.geography, year=self.year)
         }
 
     def version(self):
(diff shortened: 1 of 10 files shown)


antoniocarlon:
**Note for the acceptance:** I haven't updated the versions for the modified TIGER `ColumnTask`s. We can discuss if we want to do it now.

antoniocarlon:
Let me rethink this a bit, we should be able to use the same approach that we used in UK (not adding the year to the `Columns` task).
I recall having found a problem with that but I prefer to take a look at it again.
I'm marking this as blocked for the moment.

antoniocarlon:
I have bumped the versions of the `ColumnsTask`s.
After some research and a talk with @ethervoid, we are leaving the `year` as a parameter for the TIGER `ColumnsTask`s, so I'm unblocking the task.

tasks/us/census/acs.py
```diff
@@ -202,8 +202,8 @@ def targets(self):
     def columns(self):
         input_ = self.input()
         cols = OrderedDict([
-            ('geoidsl', input_['tiger'][self.geography + GEOID_SUMLEVEL_COLUMN]),
-            ('geoidsc', input_['tiger'][self.geography + GEOID_SHORELINECLIPPED_COLUMN]),
+            ('geoidsl', input_['tiger'][self.geography + '_{}'.format(self.year) + GEOID_SUMLEVEL_COLUMN]),
```
juanignaciosl: I'm not sure about this `self.geography + '_{}'.format(self.year)`. How come is the year now explicit but it wasn't before and nothing is removed? It's ok for 2016, but does it mean that old runs of the task didn't have `2015` in the column name but new ones would have it? Maybe it's like that but it works because "it matches" (both before the change and afterwards)?

tasks/us/census/lodes.py
```diff
@@ -563,8 +563,8 @@ def columns(self):
         data_columns = self.input()['data_meta']
         tiger_columns = self.input()['tiger_meta']
         cols = OrderedDict([
-            ('w_geocode_sl', tiger_columns['block' + GEOID_SUMLEVEL_COLUMN]),
-            ('w_geocode_sc', tiger_columns['block' + GEOID_SHORELINECLIPPED_COLUMN]),
+            ('w_geocode_sl', tiger_columns['block' + '_2015' + GEOID_SUMLEVEL_COLUMN]),
```
juanignaciosl: Same question about the year here.

juanignaciosl:
```
===== Luigi Execution Summary =====

Scheduled 39 tasks of which:
* 8 present dependencies were encountered:
    - 1 tasks.base_tasks.CreateRepoTable()
    - 1 tasks.tags.BoundaryTags()
    - 1 tasks.tags.LicenseTags()
    - 1 tasks.tags.SectionTags()
    - 1 tasks.tags.SubsectionTags()
    ...
* 31 ran successfully:
    - 1 tasks.base_tasks.Carto2TempTableTask(subdomain=observatory, table=us_landmask_union)
    - 2 tasks.base_tasks.RepoFile(...)
    - 2 tasks.simplification.ExportShapefile(schema=tiger2016, table=state, skipfailures=no) and tasks.simplification.ExportShapefile(schema=us.census.tiger, table=UnionTigerWaterGeoms_state_2016_0059278c2e, skipfailures=yes)
    - 2 tasks.simplification.SimplifyGeometriesMapshaper(...)
    - 2 tasks.simplification.SimplifyShapefile(...)
    ...

This progress looks :) because there were no failed tasks or missing external dependencies
```

PR #556, opened 2018-08-21
Add UK Postcode Districts and Sectors

antoniocarlon:
Add new UK Postcode entities (Districts and Sectors) and interpolate data for them from Output Areas.
Related to https://github.com/CartoDB/Geographica-Product-Coordination/issues/12

**Notes for the acceptance:** Execute the following tasks:
`make docker-uk-census-postcode-districts`
`make docker-uk-census-postcode-sectors`
Notice that the base branch is `551-Add_more_UK_geographies`, not `master`


juanignaciosl:
Can this acceptance be done before merging this PR? According to #554 existing data from UK should be removed. Is this needed for this acceptance as well?

antoniocarlon:
I have merged the `Shp2TempTableTask` change here, so this acceptance can be tackled as soon as the tests are green. This PR brings new geometries and data, so there is no need to remove existing UK data.

juanignaciosl:
Retest this please

tasks/simplifications.json
```diff
@@ -355,6 +355,20 @@
 		"maxmemory": "8192",
 		"skipfailures": "no"
 	},
+	"uk.odl.importpostcodedistricts_": {
```
juanignaciosl: Just out of curiosity, what does the final `_` mean?

tasks/uk/census/wrapper.py
```diff
@@ -271,11 +353,11 @@ def requires(self):
         deps = {
             'source_geom_columns': OutputAreaColumns(),
             'source_geom': OutputAreas(),
-            'source_data_columns': CensusColumns(suffix='oa'),
+            'source_data_columns': CensusColumns(),
```
juanignaciosl: Could you explain me the meaning of the change from using those two suffixes?

juanignaciosl:
Acceptance :+1:

tasks/uk/census/wrapper.py
antoniocarlon: I was using one column name for each geography level, but in the rest of DO we use the same column name for all the related geometries. From a DB schema point of view, this creates one entry in the `OBS_Column` table that relates to each `OBS_Table`

tasks/simplifications.json
antoniocarlon: It has no special meaning. That's just the `table_id` that we get from the task id using this method: https://github.com/CartoDB/bigmetadata/blob/master/tasks/base_tasks.py#L536-L543

PR #567, opened 2018-08-28
Base task for interpolations

### LINKED ISSUE #559: Extract a common pattern for data interpolation
In several tasks, we perform interpolations to get data for a geography level based on other geometries.
As the interpolations are almost always performed the same way, we can extract a common pattern an make the interpolation tasks inherit from there

antoniocarlon:
- Created two base tasks for interpolations (one for real interpolations computing the area covered by the intersection of the geometries and the other, that can be used for coupled layers, just sums up the data values of all the covered entities)
 - Refactored some tasks.

Closes #559

**Notes for the acceptance:** Execute some tasks that requires interpolations. Some eamples:
 - [x] `make -- docker-run uk.census.wrapper.CensusPostcodeSectors`
 - [x] `make -- docker-run uk.census.wrapper.CensusPostcodeDistricts`
 - [x] `make -- docker-run uk.census.wrapper.CensusPostcodeAreas`
 - [x] `make -- docker-run uk.census.wrapper.CensusLowerSuperOutputAreas`
 - [x] `make -- docker-run uk.census.wrapper.CensusMiddleSuperOutputAreas`

### DIFF
--- a/tasks/base_tasks.py
+++ b/tasks/base_tasks.py
@@ -1741,7 +1741,8 @@ class BaseInterpolationTask(TableTask):
     def requires(self):
         '''
         This method MUST be overriden in subclasses.
-        Subclasses MUST define the following requirements:
+        Subclasses MUST define the following requirements
+        (MUST return a dictionary with the following keys):
 
         'source_geom_columns'   a task with the source geometry columns
         'source_geom'           a task with the source geometries (interpolated from)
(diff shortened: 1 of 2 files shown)


tasks/base_tasks.py
```diff
@@ -1735,3 +1735,136 @@ def complete(self):
...
+        'source_geom'           a task with the source geometries (interpolated from)
+        'source_data_columns'   a task with the source data columns
+        'source_data'           a task with the source data (interpolated from)
+        'target_geom_columns'   a task with the target geometry columns
+        'target_geom'           a task with the source geometries (interpolated to)
+        'target_data_columns'   a task with the target data columns
+        '''
+        raise NotImplementedError('The requires method must be overriden in subclasses')
+
+    def get_interpolation_parameters(self):
```
juanignaciosl: Maybe documenting that it should return a dictionary with those keys would be interesting.

juanignaciosl:
Acceptance :sparkles:

PR #572, opened 2018-09-13
Parent names for geometries

juanignaciosl:
Description: CartoDB/Geographica-Product-Coordination/issues/14

@antoniocarlon could you take a quick look just to check that this is a good scaffolding, taken into account Luigi and existing classes and so on?

# Tilesets API MRLI Release

These PRs (together):
- CartoDB/bigmetadata/pull/578
- CartoDB/bigmetadata/pull/572 (on top of the previous one)
- CartoDB/observatory-extension/pull/354
- CartoDB/do_tiler/pull/51

Close these issues:
- CartoDB/bigmetadata/issues/570
- CartoDB/bigmetadata/issues/583
- CartoDB/do_tiler/issues/45


tasks/us/hierarchy.py
```diff
@@ -0,0 +1,51 @@
...
+    def _levels(self):
+        level_infos = [(level, info) for level, info in SUMLEVELS.items()]
+
+        sorted_level_infos = sorted(level_infos,
+                                    key=lambda level_info: level_info[1][
+                                        'weight'])
+        return [level_info[0] for level_info in sorted_level_infos]
+
+
+class USLevelHierarchy(Task):
```
antoniocarlon: You could extend `TempTableTask` here. Not a big deal, but it will automagically generate a `PostgresTarget` output that can be useful and won't be added to DO

tasks/us/hierarchy.py
```diff
@@ -0,0 +1,51 @@
...
+    year = IntParameter()
+    current_geography = Parameter()
+    parent_geography = Parameter()
+
+    def requires(self):
+        return {
+            'current_sumlevel': SumLevel(year=self.year,
```
antoniocarlon: Use `ShorelineClip` instead of `SumLevel` here as you will want shoreline clipped geometries

juanignaciosl:
@javitonino @antoniocarlon do you want to do a first CR? You'll probably see tons of things to change, such as the infamous `complete` overwrites ;-) Be meticulous ;-)

On the logic side of things, I've seen...
1.- Many (>75k) geometries with several parents (block < blockgroup, mainly). Is that expected?
2.- I haven't checked yet how many children don't have parents. If a child don't have any parent, you won't be able to query. In those cases I could skip levels ("if this child don't have a parent in the next level, go up until you find one"). Do you think that it's worth taking this case into account?

antoniocarlon:
Answering your questions:
 1.- I wouldn't expect so many children with several parents, especially in the US. Could that be caused by spurious intersections between different levels (intersections with a very small area caused by the minimum differences in the simplifications for example)?
 2.- I wouldn't pay much attention to this. For me is a corner case right now because we are trying to use geography levels with geometries that "match" as much as possible

antoniocarlon:
1.- I have checked a couple of intersections and I think that the intersections algorithm may be wrong (I'll dive deeper):
```
160399605001108	block	160399604002	block_group	4619.387513039036
160399605001108	block	160399605001	block_group	4619.387513039036
```


```
060770049012	block_group	06077004901	census_tract	3747872.021563051
060770049012	block_group	06077004902	census_tract	3747966.781465326
```


It seems that the child geometries lay in only one parent (I have checked that the parents don't overlap)

2.- Yes, that makes sense because `place` has not full US coverage. If we want to use all the geography levels, then we will need to do what you say (finding the next parent), but for MC I think that we don't need that, we can limit the geography levels (`place` for example is not in the MC levels)

juanignaciosl:
1.- Maybe `st_pointonsurface` might be returning a point in an edge that is within both? The "algorithm" can't be much simpler...
2.- As this is meant to work with layers with no full coverage, I'll improve the approach :+1:

requirements.txt
```diff
@@ -16,7 +16,6 @@ rst2pdf==0.93
 xlrd==1.0.0
 awscli==1.10.48
 boto==2.42.0
-pillow==3.3.0
```
javitonino: :fire:

tasks/base_tasks.py
```diff
@@ -228,18 +228,6 @@ def output(self):
...
-        '''
-        Custom complete method that attempts to check if output exists, as is
-        default, but in case of failure allows attempt to run dependencies (a
-        missing dependency could result in exception on `output`).
-        '''
-        deps = self.deps()
-        if deps and not all([d.complete() for d in deps]):
-            return False
-        else:
-            return super(TagsTask, self).complete()
```
javitonino: Wow, this can be removed? That's excellent news!

tasks/meta.py
```diff
@@ -21,11 +22,88 @@
 from sqlalchemy.types import UserDefinedType
 
 from lib.logger import get_logger
+from logging import DEBUG
 import asyncpg
 
 LOGGER = get_logger(__name__)
 
 
+class CurrentSession(object):
```
javitonino: For Python 3 the `object` inheritance is not needed. But it makes it compatible with new-style classes in Python 2. Do we still support Python 2 in `bigmetadata`?

tasks/meta.py
```diff
@@ -511,7 +589,11 @@ class OBSColumn(Base):
                                # these together across geoms: AVG, SUM etc.
 
     tables = relationship("OBSColumnTable", back_populates="column", cascade="all,delete")
-    tags = association_proxy('column_column_tags', 'tag', creator=tag_creator)
+    session = current_session()
+    # Passing the same session object to the creator saves thousands
+    # of `select 1` queries, because each use of the session was pinging
+    tag_creator_with_session = lambda x, session=session: tag_creator(x, session)
```
javitonino: The better way to do this is with [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial). Something like `tag_creator_with_session = functools.partial(tag_creator, session=session)`

antoniocarlon:
1.- Yesterday I was wrong because I used the simplified tables instead of the `observatory.obs_xxx` tables for my tests. The problem is that we are removing the geometry holes when we populate the TIGER observatory tables and thus, making the parent geometries overlap. The problem is using `ST_ExteriorRing` here https://github.com/CartoDB/bigmetadata/blob/master/tasks/us/census/tiger.py#L805-L807

I have created an issue to fix this: https://github.com/CartoDB/bigmetadata/issues/579

antoniocarlon:
Executing `make -- docker-run us.hierarchy.USLevelInclusionHierarchy --year 2015 --current-geography county --parent-geography state` raises

```
2018-09-21 08:07:40,923 [WARNING]: Will not run tasks.us.hierarchy.USLevelInclusionHierarchy(year=2015, current_geography=county, parent_geography=state) or any dependencies due to error in complete() method:
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 346, in check_complete
    is_complete = task.complete()
  File "/bigmetadata/tasks/us/hierarchy.py", line 190, in complete
    table=self.input()['level'].qualified_tablename)).fetchall()) == 0
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 626, in input
    return getpaths(self.requires())
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 826, in getpaths
    return struct.__class__((k, getpaths(v)) for k, v in six.iteritems(struct))
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 826, in <genexpr>
    return struct.__class__((k, getpaths(v)) for k, v in six.iteritems(struct))
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 824, in getpaths
    return struct.output()
  File "/bigmetadata/tasks/base_tasks.py", line 1108, in output
    self._columns = self.columns()
  File "/bigmetadata/tasks/us/census/tiger.py", line 784, in columns
    ('geoid', self.input()['geoids'][self.geography + '_{}'.format(self.year) + GEOID_SHORELINECLIPPED_COLUMN]),
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 626, in input
    return getpaths(self.requires())
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 826, in getpaths
    return struct.__class__((k, getpaths(v)) for k, v in six.iteritems(struct))
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 826, in <genexpr>
    return struct.__class__((k, getpaths(v)) for k, v in six.iteritems(struct))
  File "/usr/local/lib/python3.5/dist-packages/luigi/task.py", line 824, in getpaths
    return struct.output()
  File "/bigmetadata/tasks/base_tasks.py", line 118, in output
    for col_key, col in self.columns().items():
  File "/bigmetadata/tasks/us/census/tiger.py", line 86, in columns
    name='Shoreline clipped ' + '_{}'.format(self.year) + col.name,
AttributeError: 'NoneType' object has no attribute 'name'
```

tasks/us/hierarchy.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+def _abbreviation(geography):
+    return GEOGRAPHIES_ABBREVIATIONS.get(geography, geography)
+
+
+def abbr_tablename(target, geographies, year):
+    abbrs = [_abbreviation(geography) for geography in geographies]
+
+    if [x for x in zip(abbrs, geographies) if x[0] != x[1]]:
```
javitonino: This is hard to read (abbreviations + zip). It made me do a double take. 

Is this `if` even needed? This is to avoid the case where there are no abbreviations. Is there a problem if you replace the name even when no geography names are abbreviated?

tasks/us/hierarchy.py
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+    def target_tablename(self):
+        return abbr_tablename(super(USLevelHierarchy, self).target_tablename(),
+                              [self.current_geography, self.parent_geography],
+                              self.year)
+
+    def run(self):
+        session = current_session()
+        input = self.input()
```
javitonino: You are shadowing `input`, a built-in function :sweat_smile:

tasks/base_tasks.py
juanignaciosl: Suggested by @antoniocarlon

tasks/meta.py
juanignaciosl: I think that we haven't explicitly dropped support for Python 2, but I see no point in supporting both. What do you think, @antoniocarlon ?

tasks/targets.py
```diff
@@ -190,13 +204,11 @@ def update_or_create(self):
...
         session = current_session()
         existing = self.get(session)
-        new_version = float(self._tag.version or 0.0)
-        if existing:
-            existing_version = float(existing.version or 0.0)
-            current_session().expunge(existing)
-        else:
-            existing_version = 0.0
-        if existing and existing_version == new_version:
+        new_version = self._tag.version or 0.0
```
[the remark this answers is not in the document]
juanignaciosl: :man_shrugging:

tasks/us/census/tiger.py
```diff
@@ -892,7 +892,7 @@ class GeoNamesTable(TableTask):
     year = Parameter()
 
     def version(self):
-        return 5
+        return 4
```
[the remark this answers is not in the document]
juanignaciosl: This won't go to the final deploy. It allows me not regenerating everything. I'll mark it with a `# FIXME`.

juanignaciosl:
@antoniocarlon FYI this task has changed, but I haven't been able to reproduce it.

```
PGSERVICE=postgres10 docker-compose run -d -e LOGGING_FILE=etl_us.hierarchy.log bigmetadata luigi --module tasks.us.hierarchy tasks.us.hierarchy.USLevelInclusionHierarchy --year 2015 --current-geography county --parent-geographies '["puma", "cbsa", "congressional_district", "state"]' --log-level INFO
```

:point_up: this is just for future reference, don't try it until next CR.

tasks/us/hierarchy.py
juanignaciosl: Is there a "pythonic" convention about this? I need a short identifier (len <= 7) and I don't have any good candidate. I'll use `_input` if there's no better proposal.

antoniocarlon: It would be better to use `input_` (trailing underscore is normally used for avoiding conflict with Python keywords and leading underscore is normally used for declaring private variables)

tasks/us/hierarchy.py
juanignaciosl: I just wanted to avoid renaming tables if it's not needed. I don't have strong opinions on this.

tasks/us/hierarchy.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def requires(self):
+        levels = self._levels()
+        return {
+            'info': USHierarchyInfoUnion(year=self.year, levels=levels),
+            'rel': USHierarchyChildParentsUnion(year=self.year, levels=levels)
+        }
+
+    def _levels(self):
+        level_infos = [(level, info) for level, info in SUMLEVELS.items()]
```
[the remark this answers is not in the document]
juanignaciosl: :man_facepalming: :sweat_smile:

juanignaciosl:
Ready for next CR

tasks/meta.py
antoniocarlon: We migrated to Python 3 here https://github.com/CartoDB/bigmetadata/pull/344 so I would :fire: anything Python 2 related

juanignaciosl:
OK, :green_apple: , ready for next CR.

2da7813bc240363dc80a94666c834d6aca3f5ccf will probably break some task at hierarchy computation. I had to revert the change because it broke other stuff. I think that I had to add it because `"<schema>".Tablename` won't work if schema contains dots, but I think that we can move on and add a fix if it fails again.

tasks/targets.py
```diff
@@ -176,12 +189,16 @@ def __init__(self, tag, task):
         self._tag = tag
         self._task = task
 
+    _tag_cache = {}
+
     def get(self, session):
         '''
         Return a copy of the underlying OBSColumn in the specified session.
```
antoniocarlon: This should be `Return a copy of the underlying OBSTag in the specified session.`

tasks/meta.py
```diff
@@ -21,11 +23,88 @@
...
+        # connection
+        if self._pid != os.getpid():
+            self._session = None
+            LOGGER.debug('FORKED: {} not {}'.format(self._pid, os.getpid()))
+        if not self._session:
+            self.begin()
+        try:
+            self._session.execute("SELECT 1")
+            # Useful for debugging. This is called thousands of times
+            if LOGGER.isEnabledFor(DEBUG):
```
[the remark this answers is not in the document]
juanignaciosl: Done (although the `select 1` before is likely to make `isEnabledFor` overhead :D

antoniocarlon:
I have executed `make -- docker-run us.hierarchy.USHierarchy --year 2015` locally and it raises `Unfulfilled dependencies at run time`. Reexecuting it didn't help:

```
[ERROR]: [pid 1] Worker Worker(salt=151657519, workers=1, host=f9b76b1b349b, username=root, pid=1) failed    tasks.us.hierarchy.USHierarchyChildParentsUnion(year=2015)
  File "/usr/local/lib/python3.5/dist-packages/luigi/worker.py", line 187, in run
    raise RuntimeError('Unfulfilled %s at run time: %s' % (deps, ', '.join(missing)))
RuntimeError: Unfulfilled dependencies at run time: tasks.us.hierarchy.USLevelHierarchyWeights_block___block_group____2015_2cc12b1390, tasks.us.hierarchy.USLevelHierarchyWeights_block_group___census_tract___2015_e7749a32d6
```

antoniocarlon:
Executed locally:
```
This progress looks :) because there were no failed tasks or missing external dependencies
```
:rocket:

juanignaciosl:
Back to acceptance to finish some tasks:
- [ ] Remove `year` column.
- [ ] Generate final `tiler.us_hierarchy_child_parents` and `tiler.us_hierarchy_geonames`.
- [ ] Create `tiler.us_hierarchy_child_parents (parent_id, parent_level)` index.

antoniocarlon:
Blocked by the refactor here https://github.com/CartoDB/bigmetadata/pull/586

juanignaciosl:
Ready for CR and acceptance, together with CartoDB/do_tiler/pull/38.

In local I've generated the new table and used it from the tiler successfully.

Added `do not merge yet` because this must not be merged into master until a new DO dump is released, because it includes low level changes at ORM level that might break other stuff. As this branch is not needed for DO but for Tilesets API, this is not a problem for CR and acceptance.

scripts/parse_luigi_log.py
```diff
@@ -0,0 +1,54 @@
+#! /usr/bin/env python3
```
javitonino: Luigi has a database of executed/completed/failed tasks, so there is probably a better way to extract this information for there. However, this is simple and convenient and it's already done, so... :+1:

tasks/ca/hierarchy.py
```diff
@@ -0,0 +1,136 @@
...
+                             HierarchyChildParent, GetParentsFunction, LevelHierarchy, LevelInfo)
+from tasks.ca.statcan.geo import (GEO_PR, GEO_CD, GEO_FSA, GEO_CMA, GEO_CSD, GEO_DA, GEO_CT, GEO_DB,
+                                  Geography)
+from lib.logger import get_logger
+
+LOGGER = get_logger(__name__)
+COUNTRY = 'ca'
+
+
+def _levels():
```
javitonino: Why a function instead of a constant?

tasks/ca/hierarchy.py
```diff
@@ -0,0 +1,136 @@
...
+    return [GEO_DB, GEO_CT, GEO_DA, GEO_CSD, GEO_CMA, GEO_FSA, GEO_CD, GEO_PR]
+
+
+class CADenormalizedHierarchy(DenormalizedHierarchy):
+    @property
+    def _country(self):
+        return COUNTRY
```
javitonino: I did this for my own curiosity. It turns out you could override this with a simple variable, e.g: `_country = COUNTRY`, instead of overriding the property method.

See:
```
class Parent:
    @property
    def _prop(self):
        raise NotImplementedError('_CountryTask must define country()')
    
    def run(self):
        print(self._prop)

class Child(Parent):
    _prop = 'Hello world!'

Parent().run() -> Exception
Child().run() -> Hello world!
```

Not saying to do this, but I found it very interesting.

tasks/hierarchy.py
```diff
@@ -0,0 +1,540 @@
...
+LOGGER = get_logger(__name__)
+
+
+class _CountryTask:
+    @property
+    def _country(self):
+        raise NotImplementedError('_CountryTask must define country()')
+
+
+class DenormalizedHierarchy(Task, _CountryTask):
```
javitonino: The division of tasks in this is quite nice. It should be able to override some tasks per-country to account for differences. In particular, it should be possible to override `HierarchyChildParent` for geometries where the hierachy can be determined by the id (hierarchical ids) and save us a lot of computing time :rainbow:

tasks/mc/data.py
```diff
@@ -435,7 +435,7 @@ class AllMCData(WrapperTask):
 
     def requires(self):
         for geography in [x.replace(' ', '_') for x in GEOGRAPHIES[self.country]]:
-            yield MCData(geography=geography)
+            yield MCData(geography=geography, country=self.country)
```
javitonino: I'd change this to return the array instead of yields:
```
return [MCData(geography=x.replace(' ', '_'), country=self.country) for x in GEOGRAPHIES[self.country]]`
```

This avoids the fun thing with Luigi and yield. Example with two yields:

1. Luigi tries to execute this task
2. requires() yields T1
3. Luigi checks that T1 is not complete and executes it
4. Luigi tries to execute this task
5. requires() yields T1
6. Luigi checks that T1 is complete
7. requires() yields T2
8. Luigi checks that T2 is not complete and executes it
9. Luigi tries to execute this task
10. requires() yields T1
11. Luigi checks that T1 is complete
12. requires() yields T2
13. Luigi checks that T2 is complete
14. This task is run

Also, it inhibits parallelism (since it won't get T2 until T1 is finished. If you return an array of tasks, this goes much better:
1. Luigi tries to execute this task
2. requires() returns [T1, T2]
3. Luigi checks that T1 is not complete and executes it
4. Luigi checks that T2 is not complete and executes it (in parallel with 3 if it has multiple workers)
5. Luigi tries to execute this task
6. requires() returns [T1, T2]
7. Luigi checks that T1 is complete
8. Luigi checks that T2 is complete
9. This task is run

Less calls to complete(), less calls to requires() and potential parallelism. Do not use `yield` unless absolutely necessary (i.e: which requirements you have depend on the result of some other requirement).

https://luigi.readthedocs.io/en/stable/tasks.html#dynamic-dependencies

> This mechanism is an alternative to Task.requires in case you are not able to build up the full dependency graph before running the task. It does come with some constraints: the Task.run method will resume from scratch each time a new task is yielded.

This mentions `Task.run`, but same applies to `yields` inside `requires` (which is not even documented I think).

scripts/parse_luigi_log.py
juanignaciosl: Oh, I didn't know [that](https://luigi.readthedocs.io/en/stable/api/luigi.db_task_history.html). I'll add a comment with the link.

juanignaciosl:
```suggestion
#! /usr/bin/env python3
# This can be tackled with Luigi DB instead of parsing the log: https://luigi.readthedocs.io/en/stable/api/luigi.db_task_history.html
```

tasks/ca/hierarchy.py
juanignaciosl: The original one at US was a function and that design was propagated to other countries. But even US one can be a constant :+1: . Changing.

tasks/ca/hierarchy.py
juanignaciosl: I wouldn't change it, as overriding methods seems more natural. In addition, in the future, I think that this should be handled with configuration instead of inheritance.

tasks/mc/data.py
juanignaciosl: Same at AllMCCountries, right? Nice! Great explanation, I didn't know that and it'll be very helpful if parallelism is addressed.

javitonino: Yup, same problem there.

javitonino:
:red_circle: 

```
LOG_LEVEL=DEBUG make -- docker-run us.hierarchy.USDenormalizedHierarchy --year 2015
/usr/local/lib/python3.5/dist-packages/luigi/parameter.py:261: UserWarning: Parameter "year" with value "2015" is not of type string.
  warnings.warn('Parameter "{}" with value "{}" is not of type string.'.format(param_name, param_value))
```

It looks like we may be passing an integer (from hierarchy tasks) to ShorelineClip (which takes a string).

javitonino:
Acceptance status:

Hierachy:
- [x] US
- [ ] CA
- [ ] AU

Tilesets:
- [ ] US

juanignaciosl:
Data has been deployed to production. This will be kept unmerged until a new DO dump is released.

juanignaciosl:
This has been stuck for weeks. I'm going to merge it into `Add_more_countries_to_MC_tiler` and ask for CR and Acceptance on it.

Issue #570, opened 2018-09-10
Add remaining countries for Tilesets API


juanignaciosl:
_TBD_

juanignaciosl:
Partially? blocked by CartoDB/Geographica-Product-Coordination/issues/27

PR #574, opened 2018-09-18
Added support for more MC data

antoniocarlon:
Added support for more MC data:
 - UK
 - AU
 - CA
 - US (refactored)

Related to https://github.com/CartoDB/bigmetadata/issues/570

**Notes for the acceptance:**
 - You will need a copy of the data (the code as is expects it to be in http://172.17.0.1:8000/mc/ so you can put the `gz` files in a `mc`subdirectory and serve the parent directory doing `python -m SimpleHTTPServer`)
 - You can execute some tasks like `make -- docker-run mc.data.MCData --country au --geography state` or the whole thing doing `make -- docker-run mc.data.AllMCCountries` (it will take a while)


antoniocarlon:
I would say that the correct way would be
`PGSERVICE=postgres10 make -- docker-run mc.data.MCData --country uk --geography postcode_unit`

If this is problematic we could add some validity checks

juanignaciosl:
With that, it worked :+1: I'll check a few more countries before :rocket:

juanignaciosl:
I ran one geography for each country, making sure that it actually run (I had to remove a table from US). :rocket:

Issue #576, opened 2018-09-19
Wrong column name for geometry name in Canada 


antoniocarlon:
All the Canada geography levels have the same column for the name `ca.statcan.geo.geom_name`. This results in only one column with name `Name of Census metropolitan areas and census agglomerations` instead of the proper column name for each geography level.

Can be queried like this:
```
select *
from observatory.obs_column c, observatory.obs_column_table ct
where id like 'ca.%name%'
and c.id = ct.column_id
```

PR #580, opened 2018-09-21
Fix CA geo names

antoniocarlon:
Fixed Canada geo names
Closes https://github.com/CartoDB/bigmetadata/issues/576


javitonino:
Looks good, but got a question. How would we upgrade this? Remove all Canada metadata and re-import?

antoniocarlon:
Yes, we will need to delete some metadata to fix this. Fortunately, the new Canada 2016 data will have this from the very beginning...

javitonino:
Closing since Canada 2016 already includes this fix.

Issue #577, opened 2018-09-19
Add Australia 2016 data


antoniocarlon:
It seems that there are new geometries and new census data for Australia (2016) http://www.abs.gov.au/AUSSTATS/abs@.nsf/DetailsPage/1270.0.55.001July%202016?OpenDocument

We need to create the ETL task to import them into DO

PR #582, opened 2018-10-04
Add AU 2016 to DO

antoniocarlon:
Added Australia 2016 geographies and data to the ETL

Closes https://github.com/CartoDB/bigmetadata/issues/577

**Notes for the acceptance:**
 - AU 2011 and AU 2016 are pretty similar, so I have refactored the AU ETL task (although I have done a couple of tests, we may need to check that everything is wtill working for AU 2011)
 - The will be (for sure) some `NULL value` errors when executing all the data tasks for AU 2016 (I have only tested `STE`)


tasks/au/meta/helper2016/au_columns_helper.py
```diff
@@ -0,0 +1,205 @@
...
+        'P_Tot_75_84': 'P_Tot_75_84_yrs',
+    }
+}
+
+
+def _is_gender_column(short, gender):
+    return short.startswith('{}_'.format(gender)) or short.endswith('_{}'.format(gender))
+
+
+def _is_other_column(short, _):
```
juanignaciosl: Minor: "other"? :thinking: That's very strange :sweat_smile: Can't this be named and implemented after what it _is_, not what it's not? You can probably find something better than "other", even if it has to be "normal_measurement" name or something like that. I guess that "not-based" implementation might be related to the high number of columns. Aren't there patterns that we could match against, instead of negate a few? This kind of "not checks" often lead to surprises when new values appear.

tasks/au/meta/helper2016/au_columns_helper.py
```diff
@@ -0,0 +1,205 @@
...
+    return not short.startswith('P_') and not short.endswith('_P') \
+           and not short.startswith('M_') and not short.endswith('_M') \
+           and not short.startswith('F_') and not short.endswith('_F')
+
+
+def _filter_denominators(denominators, datapack):
+    filtered = denominators
```
juanignaciosl: Can't

```python
    filtered = denominators
    for filter_ in DENOMINATOR_FILTERS:
        filtered = list(filter(lambda x: filter_ not in x, filtered))
```

be merged into

```python
filtered = [d for d in denominators if d not in DENOMINATOR_FILTERS]
```

?

tasks/au/meta/helper2016/au_columns_helper.py
```diff
@@ -0,0 +1,205 @@
...
+                    denominators.append('Tot_P_P')
+                    denominators.append('Tot_P_F')
+                    denominators.append('P_' + short[2:])
+                elif short.endswith('_F'):
+                    denominators.append('Tot_P_P')
+                    if short != 'Tot_P_F':
+                        denominators.append('Tot_P_F')
+                    denominators.append(short[:short.rfind('_F')] + '_P')
+
+                denominators = _filter_denominators(denominators, datapack)
```
juanignaciosl: Defensive programming suggestion, feel free to ignore: as we can't have valuable tests for this kind of stuff, maybe we could check that all rows have the same number of columns, and fail otherwise.

antoniocarlon:
@juanignaciosl all the CR comments come from `au_columns_helper.py` which is just a helper script to create the `.csv` file for the 2016 columns, but not a true part of the ETL process (I have executed it once to generate the `csv` but I think that we will never execute it again).

Do you think that it's worth a refactor? I'm not saying that it's not, just asking for an unbiased opinion...

tasks/au/meta/helper2016/au_columns_helper.py
antoniocarlon: This check is true when the measurement is not a `people` one and matches for `income`, `households`, ... I'll try to find a better naming or do a small refactor

juanignaciosl:
Nah, all comments were really minor. To acceptance :rocket:

PR #585, opened 2018-10-17
Added tooling for null columns JSON files

antoniocarlon:
- Added a tool to build column restrictions JSON files from an ETL log file
 - Added a tool to merge two different column restrictions JSON files

### DIFF
--- a/tasks/tools.py
+++ b/tasks/tools.py
@@ -1,7 +1,11 @@
 import argparse
+import os
+import json
 from tasks.meta import CurrentSession
 
 REMOVE_FROM_DO = "removefromdo"
+EXTRACT_NULLS_FROM_LOG = "nullsfromlog"
+MERGE_NULLS_FILES = "mergenullsfiles"
 
 
 def remove_from_do(id):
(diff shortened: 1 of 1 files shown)


tasks/tools.py
```diff
@@ -49,12 +53,113 @@ def remove_from_do(id):
...
+
+    relevant_fields = relevant_fields.split(',')
+    cols = {}
+
+    try:
+        os.remove(outfile)
+    except OSError:
+        pass
+
+    with open(outfile, 'a') as fout:
```
javitonino: You can put multiple `with`s in a single line

```suggestion
    with open(outfile, 'a') as fout, open(infile) as fin:
```

tasks/tools.py
```diff
@@ -49,12 +53,113 @@ def remove_from_do(id):
...
+        with open(infile) as fin:
+            current_params = {}
+            for line in fin:
+                if STR_ERROR in line:
+                    current_params = {}
+                    try:
+                        vals = {item.split("=")[0].strip(): item.split("=")[1].strip()
+                                for item in line[line.rfind('(') + 1:line.rfind(')')].split(',')}
+                        for key, val in vals.items():
```
javitonino: Since `current_params` has just been emptied, I think this loop is equivalent to:

`current_params = { k: v for k, v in vals.items() if k in relevant_fields }`

tasks/tools.py
```diff
@@ -49,12 +53,113 @@ def remove_from_do(id):
...
+                    except Exception as e:
+                        pass
+
+                elif STR_NULL in line and STR_INFO not in line:
+                    columns = [x.strip() for x in line[line.rfind(STR_NULL) + len(STR_NULL):].split(',')]
+
+                    for column in columns:
+                        if column in cols:
+                            if current_params not in cols[column][STR_EXCEPTIONS]:
+                                cols[column][STR_EXCEPTIONS].extend([current_params])
```
javitonino: `extend` with an array with a single equivalent should be equivalent to `append`:

```suggestion
                                cols[column][STR_EXCEPTIONS].append(current_params)
```

?

tasks/tools.py
```diff
@@ -49,12 +53,113 @@ def remove_from_do(id):
...
+                json2 = json.load(fin2)
+
+                # The ones from infile1 that are also in infile2
+                for column1 in json1:
+                    column2 = json2.get(column1, json2.get(column1.lower(), None))
+                    if column2 is not None:
+                        exceptions1 = json1[column1][STR_EXCEPTIONS]
+
+                        for exception2 in column2[STR_EXCEPTIONS]:
+                            mergeable = _merge_item(exception2, exceptions1)
```
javitonino: I think this method is just a simple contains check:
If `item` is in `itemlist`, it returns `None`
Else, it returns the item

Furthermore, you do not care about the return type.

So, I think this can just be:
```python
if exception2 not in exceptions1:
  exceptions1.append(exception2)
```

Alternatively, if you do not care about the order, consider using `set`s (unordered list with no repeated elements), which can do this operation more efficiently and easier.

tasks/tools.py
```diff
@@ -49,12 +53,113 @@ def remove_from_do(id):
...
+
+    try:
+        os.remove(outfile)
+    except OSError:
+        pass
+
+    with open(outfile, 'a') as fout:
+        with open(infile1) as fin1:
+            json1 = json.load(fin1)
+            with open(infile2) as fin2:
```
javitonino: Same comment as above about merging `with`s

tasks/tools.py
```diff
@@ -49,12 +53,113 @@ def remove_from_do(id):
...
+                json2 = json.load(fin2)
+
+                # The ones from infile1 that are also in infile2
+                for column1 in json1:
```
javitonino: An alternative to these two loop is to iterate over all columns in both jsons and merge as needed:
```
for column in set(json1.keys() + json2.keys()):
    col1 = json1.get(column, json1.get(column.lower(), []))
    col2 = json2.get(column, json2.get(column.lower(), []))
    json1[column] = col1 + col2
```

This will generate in json1 a dictionary with all keys from json1 and json2. For keys that are in both, it will add the values without repetition.

antoniocarlon:
@javitonino thank you for the suggestions :hugs:
I think I have addressed all of them. Second CR round?

tasks/tools.py
```diff
@@ -49,12 +53,88 @@ def remove_from_do(id):
...
+
+    with open(outfile, 'a') as fout, open(infile1) as fin1, open(infile2) as fin2:
+        json1 = json.load(fin1)
+        json2 = json.load(fin2)
+
+        for column in set(list(json1.keys()) + list(json2.keys())):
+            col1 = json1.get(column, json1.get(column.lower(), {})).get(STR_EXCEPTIONS, [])
+            col2 = json2.get(column, json2.get(column.lower(), {})).get(STR_EXCEPTIONS, [])
+
+            json1[column][STR_EXCEPTIONS] = col1 + col2
```
javitonino: Does this eliminate column items that are in both jsons? Do we need to?

juanignaciosl:
Tooling FTW :heart: 

@javitonino I guess that this can be merged after your CR suggestion, right?

javitonino:
Yup, it should be ready to rock.

PR #598, opened 2018-11-08
31 us zipcodes at zooms 8 9

juanignaciosl:
This closes CartoDB/do_tiler/issues/31 together with CartoDB/do_tiler/pull/42 .

This also closes CartoDB/data/issues/7 .

Note for acceptance: as adding the logic would be too hard, before running this you must `delete from tiler.xyz_us_do_geoms where z in (8, 9)`.

### DIFF
--- a/tasks/tiler/xyz.py
+++ b/tasks/tiler/xyz.py
@@ -83,7 +83,7 @@ class TilesTempTable(Task, ConfigFile):
                        bounds Numeric[],
                        envelope Geometry,
                        ext Geometry,
-                       CONSTRAINT {table}_pk PRIMARY KEY (x,y,z)
+                       CONSTRAINT {table}_pk PRIMARY KEY (z,x,y)
                     )'''.format(schema=config['schema'],
                                 table=self._get_table_name(config))
         session.execute(sql_table)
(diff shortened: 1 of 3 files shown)


tasks/tiler/xyz.py
```diff
@@ -83,7 +83,7 @@ def _create_table(self, session, config):
                        bounds Numeric[],
                        envelope Geometry,
                        ext Geometry,
-                       CONSTRAINT {table}_pk PRIMARY KEY (x,y,z)
+                       CONSTRAINT {table}_pk PRIMARY KEY (z,x,y)
```
javitonino: Does this mean we need to rerun all existing XYZ generation and not only zooms 8 and 9?

juanignaciosl: Not at all. This is just a convenience. For the application, which always sends x, y and z, this means nothing. But, for some operations, mostly manual ones but from time to time also in the ETL (such as today's "let's change some zoom levels"), having z as the first element of the index makes some them faster.

javitonino: Then I guess we can just merge this and run in production to get the new XYZ dump (or just checkout this branch in production as acceptance+deploy).

PR #601, opened 2018-11-20
Performance fix for MC incremental

juanignaciosl:
This closes CartoDB/data/issues/7. First run worked ok with AU, but US was too slow.

### DIFF
--- a/tasks/mc/data.py
+++ b/tasks/mc/data.py
@@ -108,7 +108,7 @@ def geoname_format(country, name):
                        metro_level_name=METRO_LEVEL_NAME.get(country, 'metro'))
 
 
-HOST = 'http://172.18.0.1:8000/mc'
+HOST = 'http://172.17.0.1:8000/mc'
 # HOST = 'http://host.docker.internal:8000/mc'
 COMPLETE_URL = '{host}/carto_{country}_mrli_scores.csv.gz'
 UNTIL_URL = '{host}/carto_{country}_mrli_scores_until_{month}.csv.gz'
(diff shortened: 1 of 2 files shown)


tasks/mc/data.py
```diff
@@ -392,9 +392,14 @@ def _insert_tr(self, session):
                     FROM
                         "{input_schema}".{input_table} mc
                     WHERE mc.category = 'total retail'
-                    AND (mc.{region_id}, mc.{month}) NOT IN
-                      (select {region_id}, {month}
-                       from "{output_schema}".{output_table});
+                    AND NOT EXISTS
```
javitonino: The new query looks OK to me. Should work fine for bigger sizes as long as we have an index.

I'm curious if this is better or worse than using the `ON CONFLICT` clause for insert, e.g:
```sql
INSERT INTO wadus (region, month, fields)
SELECT mc.region, mc.month, fields
FROM input_table
ON CONFLICT (region, month) DO NOTHING
```

This moves the check to the insertion rather than selection. I guess it depends on the size of the input and output table. Assuming that output is bigger than input, it's probably worse.

tasks/mc/data.py
```diff
@@ -108,7 +108,7 @@ def geoname_format(country, name):
                        metro_level_name=METRO_LEVEL_NAME.get(country, 'metro'))
 
 
-HOST = 'http://172.18.0.1:8000/mc'
+HOST = 'http://172.17.0.1:8000/mc'
```
javitonino: Intended change?

tasks/mc/data.py
juanignaciosl: I'll give this a try, but this change makes USA be updated very fast (blocks take 2 min, while the previous code stalled at tracts), so it's not a big deal.

tasks/mc/data.py
juanignaciosl: Yes. 99% of the times Docker assigns `172.17.0.0` subnetwork.

tasks/mc/data.py
javitonino: I wrote this to feel cool (really trying to learn if the `ON CONFLICT` syntax could be used here). If the current approach is fast, this is not likely to be faster. Try if you like, but otherwise, let's merge.

juanignaciosl: Blocks, with `NOT EXISTS`: 122s
Blocks, with `ON CONFLICT` (replacing it not only in the first `INSERT` but also in the category `UPDATE`): 150s.

So, it's probably equivalent, although not faster. It probably implies row-based checks instead of a big select and filter, index-based, so it makes sense. I'll keep the `EXISTS` one.

_PS: don't merge yet, I'll also push the partial dump script for production update._

PR #611, opened 2019-01-15
Fix acs 2010 2014

javitonino:
Point to 2015 geographies to keep retrocompatibility and save disk space, even if it is wrong to do so.

### DIFF
--- a/tasks/us/census/acs.py
+++ b/tasks/us/census/acs.py
@@ -38,10 +38,6 @@ SAMPLES = [SAMPLE_5YR, SAMPLE_1YR]
 MINIMUM_TIGER_YEAR = 2015
 
 
-def tiger_year_for_acs(acs_year):
-    return str(max(MINIMUM_TIGER_YEAR, int(acs_year)))
-
-
 class DownloadACS(LoadPostgresFromZipFile):
 
     # http://censusreporter.tumblr.com/post/73727555158/easier-access-to-acs-data
@@ -81,15 +77,14 @@ class Quantiles(TableTask):
     geography = Parameter()
 
     def requires(self):
-        tiger_year = tiger_year_for_acs(self.year)
         return {
             'columns': QuantileColumns(year=self.year, sample=self.sample, geography=self.geography),
             'table': Extract(year=self.year,
                              sample=self.sample,
                              geography=self.geography),
-            'tiger': GeoidColumns(year=tiger_year),
-            'sumlevel': SumLevel(geography=self.geography, year=tiger_year),
-            'shorelineclip': ShorelineClip(geography=self.geography, year=tiger_year)
+            'tiger': GeoidColumns(year=self.tiger_year()),
+            'sumlevel': SumLevel(geography=self.geography, year=self.tiger_year()),
+            'shorelineclip': ShorelineClip(geography=self.geography, year=self.tiger_year())
         }
 
     def version(self):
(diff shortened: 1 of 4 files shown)


tasks/us/census/acs.py
```diff
@@ -35,6 +35,7 @@
                CBSA, PLACE]
 YEARS = ['2010', '2014', '2015', '2016']
 SAMPLES = [SAMPLE_5YR, SAMPLE_1YR]
+MINIMUM_TIGER_YEAR = 2015
```
juanignaciosl: _Really minor_: maybe there might be a `YEARS` array at `tiger.py` and this could get the first one. that way it'd would also document the available years.

javitonino:
> What made you dismiss that option?

In the current dump (April 2018), we have the 2010 and 2014 data bounded to 2015 geometries. This patch keeps this behaviour even if its incorrect. Main reason is that the 2014 and 2010 tiger is not at the source we use currently for tiger 2015/2016 (some S3 bucket), so including the geometries means developer effort as well as loading time and bigger dumps. Since it's something that has been broken forever, I don't think it's worth it to fix it now.

tasks/us/census/acs.py
javitonino: That's a nice idea. The years belong to tiger indeed.

juanignaciosl:
> In the current dump (April 2018), we have the 2010 and 2014 data bounded to 2015 geometries. This patch keeps this behaviour even if its incorrect

Fair enough :+1:

:rocket:

juanignaciosl:
Tests are failing, and it seems to be related:

```
tasks.meta: INFO: rollback tasks.us.zillow.Zillow_Zip_1_2011_7d584ba935: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely) (psycopg2.IntegrityError) duplicate key value violates unique constraint "obs_column_table_pkey"
DETAIL:  Key (column_id, table_id)=(us.census.tiger.zcta5_2016_geoidsl, us.zillow.zillow_zip_1_2011_7d584ba935) already exists.
 [SQL: 'INSERT INTO observatory.obs_column_table (column_id, table_id, colname) VALUES (%(column_id)s, %(table_id)s, %(colname)s)'] [parameters: ({'colname': 'region_name_sl', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.census.tiger.zcta5_2016_geoidsl'}, {'colname': 'region_name_sc', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.census.tiger.zcta5_2016_geoidsl'}, {'colname': 'allhomesplusmultifamily_zri', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.zillow.AllHomesPlusMultifamily_Zri'}, {'colname': 'singlefamilyresidencerental_zri', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.zillow.SingleFamilyResidenceRental_Zri'}, {'colname': 'singlefamilyresidence_zhvi', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.zillow.SingleFamilyResidence_Zhvi'}, {'colname': 'allhomes_zhvi', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.zillow.AllHomes_Zhvi'}, {'colname': 'allhomes_medianvaluepersqft', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.zillow.AllHomes_MedianValuePerSqft'}, {'colname': 'allhomes_medianrentalpricepersqft', 'table_id': 'us.zillow.zillow_zip_1_2011_7d584ba935', 'column_id': 'us.zillow.AllHomes_MedianRentalPricePerSqft'})]
```

Issue #613, opened 2019-01-25
ZCTA5 names


juanignaciosl:
Currently, ZCTA5 doesn't set names:

https://github.com/CartoDB/bigmetadata/blob/788fe3abc2d2d3dc5bba0a1516b5f4c82e1a5ba9/tasks/us/census/summary_levels.json#L266-L286

As they're zip codes, using the id would be nice.

- [ ] Use ids as names for zcta5.
- [ ] Rerun hierarchy names task in order to add zip codes to the hierarchy.
- [ ] Copy the geoid to the name column for US xyz table at those levels.

PR #615, opened 2019-01-30
Australia 2016

### LINKED ISSUE #608: Australia 2016 metadata problems
There are several tables with columns with the same name but different data. Since we treat them with the same column id, we get inconsistent data. An example:

The column `P_Tot_Tot` (total population) is available in several tables. Let's look at three examples:

- In table `G04 AGE BY SEX` it means the total population
- In table `G17 TOTAL PERSONAL INCOME (WEEKLY) BY AGE BY SEX ` it means total population 15 years or older
- In Table `G46 NON-SCHOOL QUALIFICATION: LEVEL OF EDUCATION(a) BY AGE BY SEX`, it means the total population 15 years or older with some qualification (aside from mandatory school). 

As you can imagine, these numbers are very different. Since for us, this column is the same, we have a mixup of data. In particular, the column that we offer depends on the database order of rows when generation `obs_meta`, so we are not even consistent. Denominators are also crazy because of this, as you can imagine. 

We need to review the generation data so:
- Each column gets a unique ID
- We only pick denominators from the same table as the original measure

alrocar:
Closes #608

I refactored a bit `au_columns_helper.py` to avoid having to update manually all the denominator aliases (sorry if the CR gets too complicated)

Also some changes in `data.py` to avoid having to process the metadata csv in the same order as it was generated.


tasks/au/data.py
```diff
@@ -193,63 +194,15 @@ def columns(self):
...
-                )
-
-                # append the rest of the subsection tags
-                col_subsections = col_subsections.split('|')
-                for subsection in col_subsections:
-                    subsection = subsection.strip()
-                    subsection_tag = subsectiontags[subsection]
-                    cols[col_id].tags.append(subsection_tag)
+        for col_id in rows:
+            self._process_col(rows[col_id], rows, session, column_reqs, cols, tablename, source, license, country, unittags, subsectiontags)
```
javitonino: It feels weird to me to pass both `rows[col_id]` and `rows` to this function. I'd either:
a) Pass `col_id` and `rows`
b) Pass `rows` and do the loop inside the function

Might be just me, I'm known for having crazy criteria for these things :P

Also, `license` is a python built-in, best to avoid using it as a variable name if possible.

tasks/au/data.py
```diff
@@ -263,6 +216,65 @@ def _fetch_requirements(self):
         with (open('{}/{}'.format(dir_path, '{}_{}_requirements.json'.format(self.profile, self.year)))) as f:
             return json.load(f)
 
+    def _process_col(self, line, rows, session, column_reqs, cols, tablename, source, license, country, unittags, subsectiontags):
+        col_id = line[1]            # B: short
+        col_name = line[2]          # C: name
+        denominators = line[3]      # D: denominators
+        col_unit = line[5]          # F: unit
```
javitonino: I like doing multi-assignment for this kind of stuff:
`_, col_id, col_name, denominator, _, col_unit = line`
but it might complicate things (you have to match exactly the number of vatiables to the array length).

Ignorable, of course

tasks/au/data.py
```diff
@@ -263,6 +216,65 @@ def _fetch_requirements(self):
...
+            reltype = 'denominator'
+            if col_agg in ['median', 'average']:
+                reltype = 'universe'
+
+            if denom_id in column_reqs:
+                targets_dict[column_reqs[denom_id].get(session)] = reltype
+            else:
+                if denom_id not in cols:
+                    # we load denominators recursively to avoid having to order
+                    # them in the source CSV file
```
javitonino: :heart:

tasks/au/data.py
```diff
@@ -263,6 +216,65 @@ def _fetch_requirements(self):
...
+            # Weight of 0 will hide this column from the user.  We generally use between 0 and 10
+            weight=5,
+            aggregate=col_agg or 'sum',
+            # Tags are our way of noting aspects of this measure like its unit, the country
+            # it's relevant to, and which section(s) of the catalog it should appear in
+            tags=[source, license, country, unittags[col_unit]],
+            targets=targets_dict
+        )
+
+        # append the rest of the subsection tags
```
javitonino: Can we do this before the OBSColumn creation and pass it in the constructor?

I'd expect something like:
`col_subsections = line[6].split('|')` to be at the top, where all other line parsing is done

`subsection_tags = [subsectiontags[subsection.strip()] for ss in col_subsections]`

`tags=[source, _license, country, unittags[col_unit]] + subsection_tags` in the constructor above

Just an idea, not sure if this is done after the constructor for any reason, but looks weird

tasks/au/geo.py
```diff
@@ -230,6 +236,7 @@ def columns(self):
         cols = OrderedDict([
             ('geom_name', geom_name),
             ('geom_id', geom_id),
+            ('external_id', external_id),
```
javitonino: Do we want to add `external_id` for all geometries? Maybe we only need to do it for one specific geometry (the exception) to avoid the duplication (and minor disk space).

Or, alternatively, just use the 11 digit code in the `geo`, and then tweak the `data` import process to calculate the 11 digit code from the 7 digit code (just a lookup in the geography table, doing partial matching).

I guess you talked about that, I just want to be up to date.

alrocar: We didn't agree on any specific solution for this. We have decided to go with this option, since it seems 7 digit codes are built from 11 digit ones (fist + six last characters):

> Or, alternatively, just use the 11 digit code in the geo, and then tweak the data import process to calculate the 11 digit code from the 7 digit code (just a lookup in the geography table, doing partial matching).

I'll let you know when it's ready for a second CR round.

Thx!

tasks/au/data.py
```diff
@@ -408,12 +414,30 @@ def populate_general(self):
...
-                      output=self.output().table,
-                      input=intable,
-                      in_colnames=', '.join(in_colnames),
-                      out_colnames='", "'.join(out_colnames))
+            if self.resolution in CONVERT_TO_7_DIGIT:
+                # we need 11 digit IDs (from the geography table) because
+                # the data table only has 7 digit IDs
+                in_colnames[0] = 'geo.geom_id'
+                cmd = '''
+                    INSERT INTO {output} ("{out_colnames}")
```
juanignaciosl: Minor: as the difference is the join, extracting the common part and adding the join afterward if needed might be more clear.

alrocar:
> Can't cols be generated completely (maybe before the for denom_id in denominators iteration)?

`OBSColumn` have to be defined for each denominators before it's added as a `target` of another `OBSColumn` so I guess we have to do recursion somehow, otherwise we'll have to iterate several times the source file and store stuff in temporal variables to finally be able to do everything in a single loop.

If you don't mind I'd go with this solution and let's see what happens with this code next time we have to load data from AU, since it's completely coupled to the source CSV files structure.

tasks/au/data.py
alrocar: In this case I prefer keep both separated. Since `in_colnames` belongs to the common part but then is changed depending on the `resolution` and both queries have different formatted variables I think it's less prone to errors (and easier to read) if both are separated.

alrocar:
Deployed to staging

tasks/au/data.py
```diff
@@ -408,12 +414,30 @@ def populate_general(self):
 
             in_colnames[0] = '"{}"'.format(self._get_geoid())
 
-            cmd = 'INSERT INTO {output} ("{out_colnames}") ' \
-                  'SELECT {in_colnames} FROM {input} '.format(
-                      output=self.output().table,
-                      input=intable,
-                      in_colnames=', '.join(in_colnames),
-                      out_colnames='", "'.join(out_colnames))
+            if self.resolution in CONVERT_TO_7_DIGIT:
```
javitonino: Shouldn't this be convert_to_11_digit? It's converting from 7 to 11 digit (completing the id), right?

alrocar: Yes, I think it makes sense

juanignaciosl:
Acceptance:

alrocar:
@juanignaciosl  Is this ready to deploy, then?

juanignaciosl:
No, no, I'll ping you when it's done. That's WIP.

juanignaciosl:
I wasn't too sure about the data check, so I did a couple of queries to `obs_column`.

This is strange, maybe a potential problem, as the only difference in the id is the case:

```
        id         |  type   |        name         |                                  description                                   | weight | aggregate | version | extra 
-------------------+---------+---------------------+--------------------------------------------------------------------------------+--------+-----------+---------+-------
 au.data.P_ToT_ToT | Numeric | Persons Total Total | Non-School Qualification:  Level of Education by Industry of Employment by Sex |      5 | sum       |       7 | 
 au.data.P_Tot_Tot | Numeric | Persons Total Total | Non-School Qualification:  Field of Study by Age by Sex                        |      5 | sum       |       7 | 
(2 rows)
```

Other than that, this is hard to test without checking all data, which is beyond my scope. Let me know what you think about that, and let see if @javitonino want to do more checks.

alrocar:
Yep, those columns seem from the previous load. We should look for columns `version = 8`, they should be prefixed with the table name so there can be duplicated totals and so on.

Also about the checks, we could do some queries through the API. I'm not sure if we can use this directly from BUILDER in staging (I cannot login to my account right now), otherwise we can just run a couple of queries to get measurements.

juanignaciosl:
I see version 8 columns, and can query some of them such as

```
gis=# WITH data AS (
  SELECT 1 AS id,
  ST_Buffer(ST_SetSRID(ST_Point(151.213, -33.88), 4326), 0.01) the_geom
),
meta AS (
  SELECT cdb_observatory.OBS_GetMeta(ST_SetSRID(ST_Extent(the_geom), 4326),
  '[{"numer_id": "au.data.G59_Tot_P",
     "normalization": "areaNormalized"}]'
  , 1, 1) meta
FROM data)
SELECT id, (data->0->>'value')::Numeric AS b01_tot_p_p_per_sq_km_2011
FROM cdb_observatory.OBS_GetData(
       (SELECT ARRAY_AGG((the_geom, id)::geomval) FROM data),
       (SELECT meta FROM meta));
 id | b01_tot_p_p_per_sq_km_2011 
----+----------------------------
  1 |           7907.75089305261
```

I don't have any more obvious check, so this is a :+1: from me. If you and @javitonino (as he opened the issue) agree, :rocket:

javitonino:
:rocket:

alrocar:
Waiting for:

- https://github.com/CartoDB/bigmetadata/issues/613
- https://github.com/CartoDB/do_tiler/issues/43

to do a joint deployment

PR #616, opened 2019-02-01
Fixes for uk

juanignaciosl:
This addresses CartoDB/bigmetadata/issues/612.

I've also made a minor fix and some integration tests at CartoDB/do_tiler/pull/59 (and added a nice-to-have).

First CR, for acceptance I'll move the data (12 and 13 from XYZ table, postcode units and sectors MC tables and UK names tables) to staging and ask for acceptance.

Code is now fixed (so running it from scratch will generate the right data), but I'll apply the fix manually on top of current data.

Export:
```sh
psql -U postgres gis --port 5555 --host 0.0.0.0 -Atc "copy (select * from tiler.xyz_uk_do_geoms where z in (12, 13)) to stdout" > ./xyz_uk_do_geoms_12_13.dump'
pg_dump -U docker -d gis --port 5555 --host 0.0.0.0 -d gis -t "\"uk.mastercard\".mc_postcode_unit" > ./mc_postcode_unit.dump
pg_dump -U docker -d gis --port 5555 --host 0.0.0.0 -d gis -t "\"uk.mastercard\".mc_postcode_sector" > ./mc_postcode_sector.dump
pg_dump -U docker -d gis --port 5555 --host 0.0.0.0 -d gis -t "tiler.uk_dz_hierarchy_geonames_2011" > ./uk_dz_hierarchy_geonames_2011.dump
```

Import:

```sh
psql -U postgres gis -Atc 'DELETE FROM tiler.xyz_uk_do_geoms WHERE z IN (12, 13);
cat /tmp/xyz_uk_do_geoms_12_13.dump | psql -U postgres -d gis -Atc "copy tiler.xyz_uk_do_geoms from stdin"

psql -U postgres gis -Atc 'DROP TABLE "uk.mastercard".mc_postcode_unit'
cat /tmp/mc_postcode_unit.dump | psql -U postgres -d gis
psql -U postgres gis -Atc 'DROP TABLE "uk.mastercard".mc_postcode_sector'
cat /tmp/mc_postcode_sector.dump | psql -U postgres -d gis
psql -U postgres gis -Atc 'GRANT USAGE ON SCHEMA "uk.mastercard" TO geocoder_api; GRANT SELECT ON ALL TABLES IN SCHEMA "uk.mastercard" TO geocoder_api;'

psql -U postgres gis -Atc 'DROP TABLE tiler.uk_dz_hierarchy_geonames_2011'
cat /tmp/uk_dz_hierarchy_geonames_2011.dump | psql -U postgres -d gis
psql -U postgres gis -Atc 'GRANT USAGE ON SCHEMA tiler TO geocoder_api; GRANT SELECT ON ALL TABLES IN SCHEMA tiler TO geocoder_api;'
```

### DIFF
--- a/tasks/mc/data.py
+++ b/tasks/mc/data.py
@@ -510,6 +510,16 @@ class MCData(TempTableTask):
         session.execute(query)
         session.commit()
 
+    def _remove_underscores(self, session, schema, table, column):
+        if self.country != 'uk' or self.geography not in ['postcode_sector', 'postcode_unit']:
+            return
+
+        session.execute('''
+            UPDATE "{schema}".{table}
+            SET {column} = REPLACE({column}, '_', ' ')
+            WHERE {column} LIKE '%\_%'
+        '''.format(schema=schema, table=table, column=column))
+
     def run(self):
         session = current_session()
         try:
@@ -519,6 +529,10 @@ class MCData(TempTableTask):
             for catid, catname in CATEGORIES.items():  # Insert the rest of the categories
                 if catid != 'TR':
                     self._update_category(session, (catid, catname))
+            self._remove_underscores(session,
+                                     self.output().schema,
+                                     self.output().tablename,
+                                     'region_id')
         except Exception as e:
             LOGGER.error('Error creating/populating {table}: {error}'.format(
                 table=self.output().table,
--- a/tasks/uk/hierarchy.py
+++ b/tasks/uk/hierarchy.py
@@ -21,7 +21,8 @@ LEVEL_CLASSES = {
     'datashare_pa_geo': PostcodeAreas
 }
 
-LEVELS = list(LEVEL_CLASSES.keys())
+# LEVEL_CLASSES.keys() can't be used because it doesn't guarantee order
+LEVELS = ['cdrc_the_geom', 'odl_ps_geo', 'gov_lsoa_geo', 'gov_msoa_geo', 'odl_pd_geo', 'datashare_pa_geo']
 
 
 def geography(level):
--- a/tasks/uk/hierarchy.py
+++ b/tasks/uk/hierarchy.py
@@ -1,4 +1,5 @@
 from lib.logger import get_logger
+from collections import OrderedDict
 from tasks.hierarchy import DenormalizedHierarchy, GetParentsFunction, \
     Hierarchy, HierarchyInfoUnion, HierarchyChildParentsUnion, \
     HierarchyChildParent, LevelHierarchy, LevelInfo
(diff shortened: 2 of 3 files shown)


tasks/mc/data.py
```diff
@@ -510,6 +510,16 @@ def _update_category(self, session, category):
         session.execute(query)
         session.commit()
 
+    def _remove_underscores(self, session, schema, table, column):
+        if self.country != 'uk' or self.geography not in ['postcode_sector', 'postcode_unit']:
+            return
+
+        session.execute('''
+            UPDATE "{schema}".{table}
```
alrocar: This question may not have much sense, but could this query be done as part of any of the other queries of the ETL?

I'm not sure how expensive is this query to be performed (maybe not too much) but just in case we could save some time 👼

tasks/uk/hierarchy.py
```diff
@@ -21,7 +21,8 @@
     'datashare_pa_geo': PostcodeAreas
 }
 
-LEVELS = list(LEVEL_CLASSES.keys())
+# LEVEL_CLASSES.keys() can't be used because it doesn't guarantee order
+LEVELS = ['cdrc_the_geom', 'odl_ps_geo', 'gov_lsoa_geo', 'gov_msoa_geo', 'odl_pd_geo', 'datashare_pa_geo']
```
antoniocarlon: You could use an `OrderedDict`

tasks/mc/data.py
juanignaciosl: I also thought about this, but I think that this is for the best. It's not expensive (seconds, or less). While doing it afterward might look cheap, it's not a problem for a performance point of view, makes it absolutely explicit, and less error-prone (if you check the code, there are several queries where ids are inserted).

@antoniocarlon , do you agree?

tasks/uk/hierarchy.py
juanignaciosl: Indeed (I realized about it afterward and forgot to fix it). On it!

tasks/mc/data.py
antoniocarlon: Given that it's executed only in certain conditions (Postcode sectors and units for the UK) I would not add it to the `insert` to keep the code cleaner there. Also if we have other exceptions in the future we can add them here, so I would keep it

juanignaciosl:
Comment honored, and I answered another one. :+1: , :-1: , @antoniocarlon , @alrocar ?

alrocar:
Assigning back to @juanignaciosl for deployment

PR #619, opened 2019-02-06
update AU 2016 columns for tiler

alrocar:
CartoDB/do_tiler#52

### DIFF
--- a/tasks/tiler/conf/au_all.json
+++ b/tasks/tiler/conf/au_all.json
@@ -1,0 +1,580 @@
+{
+    "country": "au",
+    "geolevels": {
+        "state": {
+            "geography": "au.geo.STE_2016",
+            "geoname": "au.geo.STE_name_2016",
+            "zoomlevels": [0, 1, 2, 3],
+            "simplification": 0.1
+        },
+        "sa4": {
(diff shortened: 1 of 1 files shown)


juanignaciosl:
Correction: you probably want to merge this into `Add_more_countries_to_MC_tiler`, or `608-australia-metadata`, not master. If you merge it into `608-australia-metadata` we could perform both acceptances at the same time.

_Gentle reminder: we'll push `Add_more_countries_to_MC_tiler` acceptance after the next DO dump so we get rid of this annoyance_.

juanignaciosl:
@alrocar this can be merged into `608-australia-metadata` now, right? They don't have different acceptances, and it has already been CR'ed.

alrocar:
Yes, but actually I see this PR more related to this other [issue](https://github.com/CartoDB/do_tiler/issues/52). So we can do what you prefer, either merge this to `608-australia-metadata` now, or just leave it pending until we can unblock what is needed to continue with the other issue.

juanignaciosl:
@alrocar ok, let's bound this ("the tilesets side") to that issue. Marked as blocked.

juanignaciosl:
Changed base branch because `608-australia-metadata` had already been merged.

PR #621, opened 2019-02-15
Google Storage support for Mastercard updates

juanignaciosl:
This commits allows Mastercard updates to fetch the data directly from
the bucket. Example for loading one month of Australia:

```
PGSERVICE=postgres10 docker-compose run -d -e LOGGING_FILE=etl_mc.data.log -e GOOGLE_APPLICATION_CREDENTIALS=/bigmetadata/tmp/mrli-reader.json bigmetadata luigi --module tasks.mc.data tasks.mc.data.AllMCData --country au --month 201811 --log-level INFO
```

This will be used to update the data. See CartoDB/do_tiler/issues/66.

The updated rows can be dumped with `scripts/dump_mastercard_months.sh`.

The dumped rows can be loaded with `scripts/load_mastercard_months.sh`.

For a complete automation we should:

1.- Trigger the load automatically when the bucket content changes.
2.- Run the dump for the updated months.
3.- Copy the dump files.
4.- Run the load.

Now, it's a matter of plugging the pipeline, but all tools are there :-)

For acceptance, updating all staging data would be quick and complete. I've already updated Australia.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
       - 80
 
   bigmetadata_daemon:
-    image: carto/bigmetadata
+    image: carto/bigmetadata:1.1
     volumes:
       - .:/bigmetadata
       - ${DO_LOCAL_TMP_DIR:-./tmp}:/bigmetadata/tmp
@@ -61,7 +61,7 @@ services:
     command: "luigid"
 
   bigmetadata:
-    image: carto/bigmetadata
+    image: carto/bigmetadata:1.1
     volumes:
       - .:/bigmetadata
       - ${DO_LOCAL_TMP_DIR:-./tmp}:/bigmetadata/tmp
--- a/tasks/base_tasks.py
+++ b/tasks/base_tasks.py
@@ -17,17 +17,19 @@ from collections import OrderedDict
 from datetime import date
 
 from luigi import (Task, Parameter, LocalTarget, BoolParameter, IntParameter,
-                   ListParameter, DateParameter, WrapperTask, Event)
+                   ListParameter, DateParameter, WrapperTask, Event, ExternalTask)
 from luigi.contrib.s3 import S3Target
 
 from sqlalchemy.dialects.postgresql import JSON
+
+from google.cloud import storage
 
 from lib.util import digest_file
 from lib.logger import get_logger
 
 from tasks.meta import (OBSColumn, OBSTable, metadata, current_session,
                         session_commit, session_rollback, GEOM_REF)
-from tasks.targets import (ColumnTarget, TagTarget, CartoDBTarget, PostgresTarget, TableTarget, RepoTarget)
+from tasks.targets import (ColumnTarget, TagTarget, CartoDBTarget, PostgresTarget, TableTarget, RepoTarget, URLTarget)
 from tasks.util import (classpath, query_cartodb, sql_to_cartodb_table, underscore_slugify, shell,
                         create_temp_schema, unqualified_task_id, generate_tile_summary, uncompress_file,
                         copyfile)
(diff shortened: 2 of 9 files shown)


tasks/targets.py
```diff
@@ -476,3 +476,15 @@ def exists(self):
                     function_name=self._function_name)
 
         return len(self._session.execute(query).fetchall()) > 0
+
+
+class URLTarget(Target):
+    def __init__(self, url):
+        self.value = url
```
antoniocarlon: I think that this can be removed

docker-compose.yml
```diff
@@ -61,7 +61,7 @@ services:
     command: "luigid"
 
   bigmetadata:
-    image: carto/bigmetadata
+    image: carto/bigmetadata:1.1
```
antoniocarlon: are we planning to upload to Docker Hub the new version and keep it there? If not we could just generate the new image with the existing tag and upload it when this PR is merged

tasks/base_tasks.py
```diff
@@ -1906,3 +1908,44 @@ def populate(self):
...
+        blob = bucket.blob(source_blob_name)
+
+        LOGGER.info('Downloading blob {} to {}.'.format(
+            source_blob_name,
+            destination_file_name))
+
+        blob.download_to_filename(destination_file_name)
+
+    def complete(self):
+        return os.path.exists(self._destination())
```
antoniocarlon: We should write a proper `exists` method  for `URLTarget` instead of writing a custom `complete` method for the Task

tasks/targets.py
```diff
@@ -476,3 +476,15 @@ def exists(self):
...
+        self.path = url
+
+    def exists(self):
+        return True
```
antoniocarlon: Related to https://github.com/CartoDB/bigmetadata/pull/621/files#diff-f1267b51a30fc6c52919dc99cf97448dR1948

docker-compose.yml
juanignaciosl: In fact, I already pushed it so it was available to staging. Shouldn't I have done it?

antoniocarlon: yes, no problem, but we use to update the `carto/bigmetadata` image instead of creating new versions (anyway we can do it later, so no problem)

tasks/base_tasks.py
juanignaciosl: This is the comment I was waiting for. I'm not sure about this design, but I couldn't come up with a better one without a deeper refactor. Let me explain how come did I get to a `exists` that just `return True` in `URLTarget`, because I might be wrong but it's intentional.

My problems began at `DownloadGUnzipMC`, which extends `RepoFileGUnzipTask`. That parent class forces you to implement `get_url`, so it implicitly makes an assumption: any potential "repo file" is available with a URL. That's not true for Google Storage: while you can sign files to generate URLs, they only last 7d maximum. The right way to download them is with the credentials and the SDK.

So, I changed the task so instead of assuming the URL that it needed, [it required another URL](https://github.com/CartoDB/bigmetadata/pull/621/commits/1b4319faadc413464c4a830c677e304803f48e49#diff-77ce7c6af6c29a58f760831b1f17072fR164). That URL might be a "hardcoded one" (`from URLTask`, which is just a wrapper for a value), or one provided by `GoogleStorageTask` (which happens to be a "local URL").

That's why `exists` just returns `True`: it's just a value. `URLTask` doesn't check whether the URL actually exists, and `GoogleStorageTask` returns just the value once it's run.


After this explanation, what do you think? Moving `GoogleStorageTask.complete` check for `URLTarget` might make sense, but I'm not so sure about `URLTask`.

docker-compose.yml
juanignaciosl: While we can do it later, I should revert the versioning. In general, versioning is a good practice. Why would you not want to use it here? If it's not changed, shouldn't the downloaded images be removed?

antoniocarlon: 100% agree that versioning is a good practice, I was just thinking out loud because we were not using versions until now (this doesn't mean that it was the best practice, but I guess that the reason is to avoid versioning trees).
Anyway, adding a new version of the image is something that we don't do _that often_, so I would keep the versioning (and even enforce versioning in future cases)

tasks/base_tasks.py
juanignaciosl: After a talk, I've implemented this good suggestion :muscle:

juanignaciosl:
@antoniocarlon ready for second round!

alrocar:
Acceptance 🍏 

Quite easy and handy 👍 . Next time we can work on the full automation if needed.

